PR #1, opened 2015-09-11
Initialized Django project

clintonb:
ECOM-2179

@jimabramson


clintonb:
@jimabramson feel free to merge when ready.

PR #2, opened 2015-09-15
WIP: Initial Program creation / list views.

jimabramson:
ECOM-2186

@clintonb @rlucioni  I'm working on the authentication / roles part of the ACs next, but ready for feedback on what's implemented so far.

### DIFF
--- a/programs/apps/api/models.py
+++ b/programs/apps/api/models.py
@@ -1,4 +1,0 @@
-# Models that can be shared across multiple versions of the API
-# should be created here. As the API evolves, models may become more
-# specific to a particular version of the API. In this case, the models
-# in question should be moved to versioned sub-package.
--- a/programs/apps/api/serializers.py
+++ b/programs/apps/api/serializers.py
@@ -1,4 +1,31 @@
-# Serializers that can be shared across multiple versions of the API
-# should be created here. As the API evolves, serializers may become more
-# specific to a particular version of the API. In this case, the serializers
-# in question should be moved to versioned sub-package.
+"""
+Programs API serializers.
+
+Serializers that can be shared across multiple versions of the API
+should be created here. As the API evolves, serializers may become more
+specific to a particular version of the API. In this case, the serializers
+in question should be moved to versioned sub-package.
+"""
+from django.utils.translation import ugettext as _
+from rest_framework import serializers
+
+
+from programs.apps.programs import models, constants
+
+
+class ProgramSerializer(serializers.ModelSerializer):
+    """General-purpose serializer for the Program model."""
+
+    class Meta(object):  # pylint: disable=missing-docstring
+        model = models.Program
+        fields = ('id', 'name', 'description', 'category', 'certificate_type', 'status', 'created', 'modified')
+        read_only_fields = ('id', 'created', 'modified')
+
+    def validate_status(self, status):
+        """
+        Prevent creation of Programs with a status other than UNPUBLISHED.
+        """
+        if self.instance is None and status != constants.ProgramStatus.UNPUBLISHED:
+            msg = _("When creating a Program, \"{status}\" is not a valid choice.")
+            raise serializers.ValidationError(msg.format(status=status))
+        return status
--- a/programs/apps/api/v1/tests/test_views.py
+++ b/programs/apps/api/v1/tests/test_views.py
@@ -1,0 +1,136 @@
+"""
+Tests for Programs API views (v1).
+"""
+import json
+
+import ddt
+from django.core.urlresolvers import reverse
+from django.test import TestCase
+from mock import ANY
+
+from programs.apps.programs.constants import CertificateType, ProgramCategory, ProgramStatus
+from programs.apps.programs.tests.factories import ProgramFactory
+
+
+POST_FIELDS = ("name", "description", "category", "certificate_type", "status")
+CATEGORIES = (ProgramCategory.XSERIES, )
+CERT_TYPES = (CertificateType.VERIFIED, )
+STATUSES = (ProgramStatus.UNPUBLISHED, ProgramStatus.ACTIVE, ProgramStatus.RETIRED, ProgramStatus.DELETED)
+
+
+@ddt.ddt
+class ProgramsViewTests(TestCase):
+    """
+    Tests for listing / creating Programs.
+    """
+    @staticmethod
+    def _build_post_data(**kwargs):
+        """
+        Build and return a dict representation to use for POST / create.
+        """
+        instance = ProgramFactory.build(**kwargs)
+        return {k: getattr(instance, k) for k in POST_FIELDS}
+
+    def test_list(self):
+        """
+        Verify the list includes all Programs, except those with DELETED status
+        """
+        # create one Program of each status
+        for status in STATUSES:
+            ProgramFactory(name="{} program".format(status), status=status)
+
+        response = self.client.get(reverse("api:v1:programs-list"))
+        self.assertEqual(response.status_code, 200)
+        content = json.loads(response.content)
+        self.assertEqual(len(content), 3)
+        self.assertNotIn(ProgramStatus.DELETED, set(obj["status"] for obj in content))
+
+    def test_create(self):
+        """
+        Ensure the API supports creation of Programs.
+        """
+        data = self._build_post_data()
+        response = self.client.post(reverse("api:v1:programs-list"), data=data)
+        self.assertEqual(response.status_code, 201)
+        self.assertEqual(
+            json.loads(response.content),
+            {
+                "name": data["name"],
+                "description": data["description"],
+                "category": data["category"],
+                "certificate_type": data["certificate_type"],
+                "status": data["status"],
+                "id": ANY,
+                "created": ANY,
+                "modified": ANY,
+            }
+        )
+
+    @ddt.data(*POST_FIELDS)
+    def test_missing_fields(self, field):
+        """
+        Ensure that missing fields cause validation errors if required, and create with correct defaults otherwise.
+        """
+        defaults = {
+            "description": None,
+            "certificate_type": None,
+            "status": ProgramStatus.UNPUBLISHED,
+        }
+
+        data = self._build_post_data()
+        del data[field]
+        if field in defaults:
+            expected_status = 201
+        else:
+            expected_status = 400
+
+        response = self.client.post(reverse("api:v1:programs-list"), data=data)
+        self.assertEqual(response.status_code, expected_status)
+        content = json.loads(response.content)
+        if expected_status == 201:
+            self.assertEqual(content[field], defaults[field])
+        else:
+            self.assertIn("field is required", content[field][0])
+
+    @ddt.data(ProgramStatus.ACTIVE, ProgramStatus.RETIRED, ProgramStatus.DELETED, None, "", "unrecognized")
+    def test_create_with_invalid_status(self, status):
+        """
+        Ensure that it is not allowed to create a Program with a status other than "unpublished"
+        """
+        data = self._build_post_data(status=status)
+        response = self.client.post(reverse("api:v1:programs-list"), data=data)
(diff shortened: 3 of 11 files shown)


programs/apps/api/models.py
```diff
@@ -1,4 +1,65 @@
...
+in question should be moved to versioned sub-package.
+"""
+from django.db import models
+from django_extensions.db.models import TimeStampedModel
+
+from programs.apps.api import constants
+
+
+# TODO complete model docstring
+# TODO internationalize help text?   Or get rid of it!
```
clintonb: Internationalize.

programs/apps/api/models.py
```diff
@@ -1,4 +1,65 @@
...
+        unique=True,
+    )
+
+    description = models.TextField(
+        help_text='A full-length description of the Program.',
+        null=True,
+        blank=True,
+    )
+
+    category = models.CharField(
```
clintonb: Looking forward, what are the other categories/types of programs?

programs/apps/api/models.py
```diff
@@ -1,4 +1,65 @@
...
+        help_text='The category / type of Program.',
+        max_length=32,
```
clintonb: Why the limitation? Is this user-facing?

programs/apps/api/models.py
```diff
@@ -1,4 +1,65 @@
...
+        choices=_choices(constants.ProgramCategory.XSERIES),
+    )
+
+    certificate_type = models.CharField(
+        help_text='Optional certification criteria for Course Runs associated with this Program.',
```
clintonb: Lowercase course runs.

programs/apps/api/models.py
```diff
@@ -1,4 +1,65 @@
...
+        max_length=32,
+        choices=["None", None] + _choices(constants.CertificateType.VERIFIED),
```
clintonb: Mark `"None"` for translation.

programs/apps/api/models.py
```diff
@@ -1,4 +1,65 @@
-# Models that can be shared across multiple versions of the API
-# should be created here. As the API evolves, models may become more
-# specific to a particular version of the API. In this case, the models
-# in question should be moved to versioned sub-package.
+"""
```
clintonb: I should probably get rid of `api/models.py`. The odds of there being API-specific are pretty low. This file belongs in `apps/programs/models.py`, where `apps/programs` contains the logic for Programs.

programs/apps/api/serializers.py
```diff
@@ -1,4 +1,30 @@
...
+from rest_framework import serializers
+
+from programs.apps.api import constants, models
+
+
+# TODO add back URLs
+class ProgramSerializer(serializers.ModelSerializer):
+    """General-purpose serializer for the Program model."""
+
+    class Meta:  # pylint: disable=missing-docstring, old-style-class
```
clintonb: Use `class Meta (object)` to get around the pylint warning.

programs/apps/api/serializers.py
```diff
@@ -1,4 +1,30 @@
...
+        model = models.Program
+        fields = ('id', 'name', 'description', 'category', 'certificate_type', 'status', 'created', 'modified')
+        read_only_fields = ('id', 'created', 'modified')
+
+    def validate_status(self, status):
+        """
+        Prevent creation of Programs with a status other than UNPUBLISHED.
```
clintonb: What about editing? Is this limitation/validation temporary?

programs/apps/api/v1/tests/test_views.py
```diff
@@ -0,0 +1,159 @@
...
+from programs.apps.api.models import Program
+
+
+POST_FIELDS = ("name", "description", "category", "certificate_type", "status")
+CATEGORIES = (ProgramCategory.XSERIES, )
+CERT_TYPES = (CertificateType.VERIFIED, )
+STATUSES = (ProgramStatus.UNPUBLISHED, ProgramStatus.ACTIVE, ProgramStatus.RETIRED, ProgramStatus.DELETED)
+
+
+class ProgramFactory(factory.django.DjangoModelFactory):  # pylint: disable=missing-docstring
```
clintonb: Move to `apps/programs/factories.py`.

clintonb:
`ddt` needs to be added to `requirements/test.txt`.

programs/apps/api/v1/urls.py
```diff
@@ -1,3 +1,11 @@
...
 
-urlpatterns = []
+from django.conf.urls import url
+from programs.apps.api.v1 import views
+
+# Wire up our API using automatic URL routing.
+# Additionally, we include login URLs for the browsable API.
+urlpatterns = [
+    url(r'^programs/', views.ProgramsView.as_view(), name='programs-list'),
+    # url(r'^api-auth/', include('rest_framework.urls', namespace='rest_framework'))
```
clintonb: Delete.

programs/apps/api/v1/views.py
```diff
@@ -1 +1,17 @@
...
+"""
+from rest_framework import generics
+
+from programs.apps.api.constants import ProgramStatus
+from programs.apps.api.models import Program
+from programs.apps.api.serializers import ProgramSerializer
+
+
+# TODO complete docstring
+class ProgramsView(generics.ListCreateAPIView):
```
clintonb: `class ProgramsView(mixins.CreateModelMixin, mixins.ListModelMixin,  viewsets.GenericViewSet)`

I'd like to get in the habit of using [`Viewsets`](http://www.django-rest-framework.org/api-guide/viewsets/) as they clean up some boiler plate view and url code. Also, if/when we later add the retrieve/delete actions, we need only add a mixin and tests.

programs/apps/api/v1/tests/test_views.py
```diff
@@ -0,0 +1,159 @@
...
+        data = ProgramFactory.build_post_data(certificate_type=certificate_type)
+        expected_status = 400 if certificate_type == "unrecognized" else 201
+        response = self.client.post(reverse("api:v1:programs-list"), data=data)
+        self.assertEqual(response.status_code, expected_status)
+        if expected_status == 400:
+            content = json.loads(response.content)
+            self.assertIn("not a valid choice", content["certificate_type"][0])
+
+    @ddt.data(None, "", ProgramCategory.XSERIES, "unrecognized")
+    def test_create_with_category(self, category):
```
clintonb: Split this into two tests: one for the accepted category, another for the unacceptable values. That should help clean up the if-statements/branching.

programs/apps/api/v1/tests/test_views.py
```diff
@@ -0,0 +1,159 @@
...
+        data = ProgramFactory.build_post_data(status=status)
+        expected_status = 201 if status == ProgramStatus.UNPUBLISHED else 400
+        response = self.client.post(reverse("api:v1:programs-list"), data=data)
+        self.assertEqual(response.status_code, expected_status)
+        if expected_status == 400:
+            content = json.loads(response.content)
+            self.assertIn("not a valid choice", content["status"][0])
+
+    @ddt.data(None, "", CertificateType.VERIFIED, "unrecognized")
+    def test_create_with_certificate_type(self, certificate_type):
```
clintonb: Same comment as `test_create_with_category`: split into two tests.

programs/apps/api/v1/tests/test_views.py
```diff
@@ -0,0 +1,159 @@
...
+        return {k: getattr(instance, k) for k in POST_FIELDS}
+
+
+@ddt.ddt
+class ProgramsViewTests(TestCase):
+    """
+    Tests for listing / creating Programs.
+    """
+
+    def setUp(self):
```
clintonb: Delete.

programs/apps/api/v1/tests/test_views.py
```diff
@@ -0,0 +1,159 @@
...
+            ProgramFactory(name="{} program".format(status), status=status)
+
+        response = self.client.get(reverse("api:v1:programs-list"))
+        self.assertEqual(response.status_code, 200)
+        content = json.loads(response.content)
+        self.assertEqual(len(content), 3)
+        self.assertNotIn(ProgramStatus.DELETED, set(obj["status"] for obj in content))
+
+    def test_create(self):
+        """
```
clintonb: Verify the API supports the creation of Programs.

programs/apps/api/v1/tests/test_views.py
```diff
@@ -0,0 +1,159 @@
...
+
+        data = ProgramFactory.build_post_data()
+        del data[field]
+        if field in defaults:
+            expected_status = 201
+        else:
+            expected_status = 400
+
+        response = self.client.post(reverse("api:v1:programs-list"), data=data)
+        print field, response.content
```
clintonb: Delete.

programs/apps/api/models.py
jimabramson: None.  This is somewhat provisional; since xseries will be an edx-specific category this should be made a setting or plugin (using stevedore or similar).  For now I'm keeping it simple.

programs/apps/api/models.py
jimabramson: Quasi-arbitrary, but I think it's good practice to define a max length for UI/presentation sanity.

programs/apps/api/serializers.py
jimabramson: unpublished is the only valid initial state.  transitions to successive states are expected to happen after Programs are fully built up, which is out of scope for this initial set of functionality.

programs/apps/api/v1/tests/test_views.py
jimabramson: OK, gonna make that apps/programs/tests/factories.py tho.

programs/apps/api/serializers.py
```diff
@@ -1,4 +1,32 @@
...
+        """
+        if self.instance is None and status != constants.ProgramStatus.UNPUBLISHED:
+            msg = _("When creating a Program, \"{}\" is not a valid choice.")
```
rlucioni: To make this easier on translators, use a keyword as a placeholder. Also, you can print literal curly braces by doubling them up:
```python
msg = _("When creating a Program, {{status}} is not a valid choice.")
```

jimabramson:
@rlucioni addressed your comment.

@rlucioni @clintonb i'd like to go ahead and merge this, then set up a separate follow-up PR for auth functionality.  If you're down, give me your comments or thumbs up...

programs/apps/api/serializers.py
```diff
@@ -1,4 +1,32 @@
...
+Serializers that can be shared across multiple versions of the API
+should be created here. As the API evolves, serializers may become more
+specific to a particular version of the API. In this case, the serializers
+in question should be moved to versioned sub-package.
+"""
+from django.utils.translation import ugettext as _
+from rest_framework import serializers
+
+
+# TODO add back URLs
```
clintonb: Does this still need to be done?

programs/apps/programs/constants.py
```diff
@@ -0,0 +1,24 @@
+"""
```
clintonb: Use `from __future__ import unicode_literals` to avoid having to mark all strings as Unicode.

programs/apps/programs/tests/factories.py
```diff
@@ -0,0 +1,18 @@
...
+Factories for tests of Programs.
+"""
+import factory
+
+from programs.apps.programs.models import Program
+from programs.apps.programs.constants import CertificateType, ProgramCategory, ProgramStatus
+
+
+class ProgramFactory(factory.django.DjangoModelFactory):  # pylint: disable=missing-docstring
+    class Meta:  # pylint: disable=missing-docstring, old-style-class
```
clintonb: Use `class Meta(object)` to get rid of the pylint old-style-class violation.

clintonb:
:+1:

rlucioni:
:+1: once commits are squashed.

PR #5, opened 2015-09-28
Added JWT authentication, initial roles and permissions

jimabramson:
@clintonb @rlucioni Please review.
@ormsbee @nasthagiri FYI.

This adds JWT authentication (and only JWT authentication) for the REST API.  A little redirection trick when using the oidc login via browser makes the browseable API usable enough (though it doesn't handle refreshing expired tokens).

I have decided to leave the user model in place, even though it is theoretically not needed.  The basic reasoning is that it helps to use django's APIs for groups and permissions, and perhaps more significantly, doesn't require a major overhaul of the social-auth pipeline / oidc client, which is a piece I've found rather difficult to debug. 

At present one can only login to the API if a properly configured LMS is running to provision the id_tokens.  In a separate PR we can modify auto-auth to generate self-issued tokens with long expirations which should make various kinds of manual/automated testing easier.


programs/apps/api/authentication.py
```diff
@@ -0,0 +1,66 @@
...
+
+    def authenticate_credentials(self, payload):
+        """
+        Return a user object to be associated with the present request, based on
+        the content of an already-decoded / verified JWT payload.
+
+        In the process of inflating the user object based on the payload, we also
+        make sure that the roles associated with this user are up-to-date.
+        """
+        username = payload['preferred_username']
```
rlucioni: Use `get()` here to avoid KeyErrors. If the key is missing, raise DRF's `AuthenticationFailed` exception.

programs/apps/api/authentication.py
```diff
@@ -0,0 +1,66 @@
...
+    def get_jwt_value(self, request):
+        """
+        See if we can get the JWT from the query string, before checking HTTP headers.
+
+        This is useful for using the DRF browseable API interface.  However it is a
+        security hazard if used without HTTPS, so we should assume that the application
+        is never deployed without HTTPS in a public environment.
+        """
+        if request.method == 'GET' and 'id_token' in request.GET:
+            return request.GET['id_token']
```
rlucioni: Consider using `.query_params` over `.GET`. For context, see http://www.django-rest-framework.org/api-guide/requests/#query_params.

programs/apps/api/tests/test_authentication.py
```diff
@@ -0,0 +1,65 @@
...
+        with mock.patch(
+            'auth_backends.backends.EdXOpenIdConnect.validate_and_return_id_token',
+            return_value=id_token_payload,
+        ):
+            response = self.client.get(
+                reverse('social:complete', args=['edx-oidc']),
+                data={'state': 'test-oauth-state', 'code': 'dummy-code'},
+            )
+            self.assertEqual(response.status_code, 302)
+            self.assertEqual(response['Location'], expected_redirect_uri)
```
rlucioni: You can replace these two lines with a call to [`assertRedirects()`](https://docs.djangoproject.com/en/1.8/topics/testing/tools/#django.test.SimpleTestCase.assertRedirects). For example, see [here](https://github.com/edx/ecommerce/blob/master/ecommerce/tests/test_urls.py#L26).

programs/apps/api/v1/tests/mixins.py
```diff
@@ -0,0 +1,46 @@
...
+        """Generate a JWT token with the provided payload."""
+        secret = secret or self.JWT_SECRET_KEY
+        token = jwt.encode(payload, secret)
+        return token
+
+    def generate_id_token(self, user, admin=False, ttl=0):
+        """Generate a JWT id_token that looks like the ones currently
+        returned by the edx oidc provider."""
+
+        now = int(datetime.now().strftime('%s'))
```
rlucioni: If all you're after here is the time in seconds since the epoch, you could do:
```python
from time import time

now = int(time())
```

programs/apps/api/v1/tests/mixins.py
```diff
@@ -0,0 +1,46 @@
...
+            "iss": self.JWT_ISSUER,
+            "sub": user.pk,
+            "aud": self.JWT_AUDIENCE,
+            "nonce": "dummy-nonce",
+            "exp": now + ttl,
+            "iat": now,
+            "preferred_username": user.username,
+            "administrator": admin,
+            "email": user.email,
+            "locale": "en",
```
rlucioni: Should this be pulled from Django settings?

rlucioni: Whoops, ignore me. I somehow missed that this was a testing mixin.

programs/settings/base.py
```diff
@@ -168,8 +205,16 @@
...
-# TODO Set this to another (non-staff, ideally) path.
-LOGIN_REDIRECT_URL = '/admin/'
+LOGIN_REDIRECT_URL = '/api/v1/programs/'
+
+JWT_AUTH = {
+    'JWT_SECRET_KEY': None,
+    'JWT_ALGORITHM': 'HS256',
+    'JWT_VERIFY_EXPIRATION': True,
+    'JWT_ISSUER': None,
+    'JWT_PAYLOAD_GET_USERNAME_HANDLER': lambda d: d.get('preferred_username'),
```
rlucioni: Why is the username keyed under "preferred_username" in the payload, instead of just "username"?

programs/settings/local.py
```diff
@@ -48,10 +48,13 @@
...
 
 # AUTHENTICATION
+
+OAUTH2_PROVIDER_URL = 'http://127.0.0.1:8000/oauth2'
+
 # Set these to the correct values for your OAuth2/OpenID Connect provider (e.g., devstack)
 SOCIAL_AUTH_EDX_OIDC_KEY = 'replace-me'
-SOCIAL_AUTH_EDX_OIDC_SECRET = 'replace-me'
-SOCIAL_AUTH_EDX_OIDC_URL_ROOT = 'replace-me'
+SOCIAL_AUTH_EDX_OIDC_SECRET = SOCIAL_AUTH_EDX_OIDC_SECRET
```
rlucioni: Why bother with this? `SOCIAL_AUTH_EDX_OIDC_SECRET` is imported from the base settings.

programs/settings/base.py
jimabramson: this is how the id_tokens provided by our oidc provider are structured, e.g. https://github.com/edx/auth-backends/blob/master/auth_backends/backends.py#L27, and is the standard claim name under the openid connect spec: http://openid.net/specs/openid-connect-core-1_0.html#StandardClaims

rlucioni: Got it, thanks.

programs/settings/local.py
jimabramson: good catch, removed.

jimabramson:
@rlucioni thanks for feedback - addressed your comments in a fixup.

rlucioni:
:+1:

programs/apps/api/v1/views.py
```diff
@@ -13,5 +15,7 @@ class ProgramsViewSet(mixins.CreateModelMixin, mixins.ListModelMixin, viewsets.G
     """
     List / Create Programs.
     """
+    authentication_classes = (JwtAuthentication, )
```
clintonb: Make these the default authentication class and permissions in settings.

programs/settings/local.py
```diff
@@ -48,10 +48,12 @@
 # END TOOLBAR CONFIGURATION
 
 # AUTHENTICATION
+
+OAUTH2_PROVIDER_URL = 'http://127.0.0.1:8000/oauth2'
+
 # Set these to the correct values for your OAuth2/OpenID Connect provider (e.g., devstack)
 SOCIAL_AUTH_EDX_OIDC_KEY = 'replace-me'
-SOCIAL_AUTH_EDX_OIDC_SECRET = 'replace-me'
```
clintonb: Why no secret?

clintonb:
Why not use session authentication for the browseable API?

programs/settings/local.py
jimabramson: its value would be equal to that defined in base.py - in fact i think _KEY should go too, for the same reason.

furthermore, there's a lot of overlap / cross-assignment of the settings for SOCIAL_AUTH_EDX_*, OAUTH2_*, and JWT_AUTH which is pretty confusing at this point.  I'll look at making some simplifications.

programs/apps/api/v1/views.py
jimabramson: good idea, done

jimabramson:
@clintonb main reason to disable session auth is to avoid the need for CSRF protection.

jimabramson:
@clintonb updated with minor improvements to authentication settings.

clintonb:
@jimabramson what's the harm in CSRF? What, if anything, do we gain from trying to work around de-facto standard behavior?

jimabramson:
@clintonb do you have a suggestion for how to pass / use csrf tokens when the client is using JWT authentication?

clintonb:
Session authentication is only used for the browseable API, and all of
those pages already know how to handle CSRF tokens. You can continue using
JWT authentication for all other requests.

-- 
Clinton Blackburn

jimabramson:
@clintonb last fixup adds session authentication and removes the query string / redirection hackery.  i've verified that requests using session auth require a csrf token, but jwt auth requests do not.  hooray, less code.  thanks for the help!

programs/apps/api/v1/views.py
```diff
@@ -44,5 +47,7 @@ class ProgramsViewSet(mixins.CreateModelMixin, mixins.ListModelMixin, viewsets.G
         * modified: The date/time this Program was last modified.
 
     """
+    authentication_classes = (JwtAuthentication, SessionAuthentication)
```
clintonb: Move both of the authentication and permission declarations to the defaults in settings. The (very few?) classes that need to override them can, but I'd imagine almost all of the endpoints will use the same authentication and permissions.

programs/apps/programs/migrations/0003_auto_20150921_1944.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+def create_groups(apps, schema_editor):
+
+    Group = apps.get_model('auth', 'Group')
+    Permission = apps.get_model('auth', 'Permission')
+    ContentType = apps.get_model('contenttypes', 'ContentType')
+
+    learner_group, created = Group.objects.get_or_create(name=Role.LEARNERS)
+    if created:
```
clintonb: This isn't needed since it's a no-op.

programs/apps/programs/migrations/0003_auto_20150921_1944.py
```diff
@@ -0,0 +1,48 @@
...
+        pass
+
+    author_group, created = Group.objects.get_or_create(name=Role.AUTHORS)
+    if created:
```
clintonb: This isn't needed since it's a no-op.

programs/apps/programs/migrations/0003_auto_20150921_1944.py
```diff
@@ -0,0 +1,48 @@
...
+        pass
+
+    admin_group, created = Group.objects.get_or_create(name=Role.ADMINS)
+    if created:
+        Program = apps.get_model('programs', 'Program')
+        content_type = ContentType.objects.get_for_model(Program)
+
+        perm, __ = Permission.objects.get_or_create(codename='add_program', content_type=content_type)
```
clintonb: There is an easier way to create custom permissions: https://courses.stage.edx.org/api/credit/v1/courses/.

Also, the add permission should have been created when the model was created, along with the change and delete permissions.

jimabramson:
@clintonb cleaned up.

programs/apps/programs/migrations/0003_auto_20150921_1944.py
jimabramson: @clintonb i added a comment in the last commit explaining why the permission is being created explicitly during the migration.

clintonb:
:+1:

PR #6, opened 2015-09-28
WIP API documentation

jimabramson:
@clintonb @peter-fogg @bderusha @rlucioni 

establishing api documentation, following the guideline: https://openedx.atlassian.net/wiki/display/AC/edX+REST+API+Conventions#edXRESTAPIConventions-8.Documentation

The swagger output looks nice, though it's not perfect.
The inability to create separate documentation sections for each method in a viewset is troublesome, wonder if anyone knows a workaround that doesn't involve restructuring code.

### DIFF
--- a/programs/apps/api/v1/views.py
+++ b/programs/apps/api/v1/views.py
@@ -8,10 +8,41 @@ from programs.apps.programs.models import Program
 from programs.apps.api.serializers import ProgramSerializer
 
 
-# TODO complete docstring
 class ProgramsViewSet(mixins.CreateModelMixin, mixins.ListModelMixin, viewsets.GenericViewSet):
     """
-    List / Create Programs.
+
+    **Use Cases**
+
+        List existing Programs, and create new ones.
+
+    **Example Requests**
+
+        # Return a list of programs in the system.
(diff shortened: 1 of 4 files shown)


clintonb:
:+1:

programs/apps/api/v1/views.py
```diff
@@ -8,10 +8,41 @@
...
+
+        Only users with global administrative rights may create programs.  POST requests from non-
+        admins will result in status 403.
+
+    **Response Values**
+
+        * id: The ID of the program.
+        * name: The user-facing display name for this Program.
+        * description: A full-length description of the Program.
+        * category: The category / type of Program.  Right now the only value allowed in 'xseries'.
```
rlucioni: s/in/is

rlucioni:
@jimabramson from the DRF site's [section](http://www.django-rest-framework.org/topics/documenting-your-api/#endpoint-documentation) about documentation:

> Note that one constraint of using viewsets is that any documentation be used for all generated views, so for example, you cannot have differing documentation for the generated list view and detail view.

Looks good aside from a tiny typo, :+1: .

PR #8, opened 2015-10-02
move API view permissions to defaults.

jimabramson:
@clintonb recommended this yesterday after #5 was merged.  no functionality is changed.


clintonb:
:+1:

PR #9, opened 2015-10-07
Implement program view, course code / org lists, and relational models.

jimabramson:
ECOM-2188

@clintonb @rlucioni please review.  @cpennington FYI (comments welcome).

Summarizing what's in here:
* All models (except for ProgramUser) implemented.
* Added detail GET for Programs.
* Program serializer now does inline serialization for Organization, CourseCode, and CourseRunMode.
* Pagination for all the list endpoints.
* Added list filters for programs and course code endpoints.
* Moving away from django model permissions and using DRF permission backends instead.
* Minor refactoring of authentication, to deal with issues encountered using session auth with browseable API.


programs/apps/api/permissions.py
```diff
@@ -0,0 +1,33 @@
...
+            super(IsAdminGroupOrReadOnly, self).has_permission(request, view) and
+            request.method in permissions.SAFE_METHODS or
+            request.user.groups.filter(name=Role.ADMINS).exists()
+        )
+
+
+class IsAdminGroup(permissions.IsAuthenticated):
+    """
+    Allow read-only access for any authenticated user, but require membership
+    in the ADMINS group for writes.
```
jimabramson: docstring is wrong, will fix locally.

programs/apps/api/filters.py
```diff
@@ -0,0 +1,59 @@
...
+    def filter_queryset(self, request, queryset, view):
+        if request.method == 'GET' and 'status' in request.query_params:
+            return queryset.filter(status=request.query_params['status'])
+        else:
+            return queryset
+
+
+class ProgramOrgKeyFilterBackend(filters.BaseFilterBackend):
+    """
+    Allows for filtering programs or course codes by an organization key query string argument.
```
jimabramson: will fix

programs/apps/api/filters.py
```diff
@@ -0,0 +1,59 @@
...
+    def filter_queryset(self, request, queryset, view):
+        if request.method == 'GET' and 'organization' in request.query_params:
+            return queryset.filter(organizations__key=request.query_params['organization'])
+        else:
+            return queryset
+
+
+class CourseCodeOrgKeyFilterBackend(filters.BaseFilterBackend):
+    """
+    Allows for filtering programs or course codes by an organization key query string argument.
```
jimabramson: will fix

programs/apps/api/pagination.py
```diff
@@ -0,0 +1,26 @@
+"""
+Reusable pagination for REST API views.
+"""
+from rest_framework.response import Response
+from rest_framework import pagination
+
+
+class DefaultPagination(pagination.PageNumberPagination):
```
cpennington: Does this suffer from the same pagination issues that we're seeing on Ecommerce? https://openedx.atlassian.net/browse/ECOM-2501

programs/apps/api/filters.py
```diff
@@ -0,0 +1,59 @@
...
+from programs.apps.programs.constants import ProgramStatus
+
+
+class ProgramStatusRoleFilterBackend(filters.BaseFilterBackend):
+    """
+    Depending on the group membership of the requesting user, filter program
+    results according to their status:
+
+      - ADMINS can see programs with any status other than 'deleted'
+      - LEARNERS can see programs with any stutus other than 'deleted' or 'unpublished'
```
rlucioni: s/stutus/status

programs/apps/api/filters.py
```diff
@@ -0,0 +1,59 @@
...
+class CourseCodeOrgKeyFilterBackend(filters.BaseFilterBackend):
+    """
+    Allows for filtering programs or course codes by an organization key query string argument.
+    """
+
+    def filter_queryset(self, request, queryset, view):
+        if request.method == 'GET' and 'organization' in request.query_params:
+            return queryset.filter(organization__key=request.query_params['organization'])
+        else:
+            return queryset
```
rlucioni: These filter backends contain a lot of duplicated code. Can you cut down on that with a shared helper, potentially mixed into the `BaseFilterBackend`?

The `ProgramOrgKeyFilterBackend` and `CourseCodeOrgKeyFilterBackend` also look exactly the same. Is that correct?

programs/apps/api/pagination.py
jimabramson: It might.  @clintonb suspected a missing index in that ticket, which I think is a plausible explanation, and if so, the same remedy would apply here.  otherwise there's something deeper inside django / DRF that's causing trouble, as we aren't doing anything custom here.

programs/apps/api/filters.py
jimabramson: they are one (important) character different - see L#45 & L#57.  in prog org the organization relationship is a m2m so we can't reuse.  However, i'm sure we can DRY this module off somewhat; I will propose a refactor in a fixup commit.

rlucioni: I see it, thanks.

programs/apps/api/v1/tests/test_views.py
```diff
@@ -4,26 +4,36 @@
...
+    ProgramOrganizationFactory,
+)
 
 
 USERNAME = 'preferred_username'
-POST_FIELDS = ("name", "description", "category", "status")
+POST_FIELDS = ("name", "subtitle", "category", "status")
 CATEGORIES = (ProgramCategory.XSERIES, )
 STATUSES = (ProgramStatus.UNPUBLISHED, ProgramStatus.ACTIVE, ProgramStatus.RETIRED, ProgramStatus.DELETED)
+DRF_DATE_FORMAT = u'%Y-%m-%dT%H:%M:%S.%fZ'
```
rlucioni: Is it possible to use `isoformat()` instead of this format string? I understand the need for the format string if DRF uses one like it to serializes datetimes, but I'm surprised it's not just using `isoformat()` behind the scenes.

programs/apps/api/v1/tests/test_views.py
```diff
@@ -82,13 +144,108 @@ def test_create(self):
...
+                u"id": ANY,
+                u"created": ANY,
+                u"modified": ANY,
+            }
+        )
+
+    @ddt.data(*STATUSES)
+    def test_view_admin(self, status):
+        """
+        """
```
rlucioni: Could you fill in this format string and the one below?

programs/apps/api/v1/tests/test_views.py
```diff
@@ -149,3 +306,122 @@ def test_create_duplicated_name(self):
...
+        Ensure the API supports creation of Organizations.
+        """
+        data = {'key': 'edX', 'display_name': 'edX University'}
+        client = APIClient()
+        client.force_authenticate(user=self._make_user(Role.ADMINS))
+        response = client.post(reverse("api:v1:organizations-list"), data)
+        self.assertEqual(response.status_code, 201)
+        self.assertEqual(
+            json.loads(response.content),
```
rlucioni: Since you're using DRF's `APIClient` here, you can inspect `response.data` instead of parsing `response.content`.

programs/apps/api/v1/tests/test_views.py
```diff
@@ -149,3 +306,122 @@ def test_create_duplicated_name(self):
...
+        """
+        Ensure the API supports listing of Organizations by admins.
+        """
+        for _ in range(3):
+            OrganizationFactory.create()
+        client = APIClient()
+        client.force_authenticate(user=self._make_user(Role.ADMINS))
+        response = client.get(reverse("api:v1:organizations-list"))
+        self.assertEqual(response.status_code, 200)
+        results = json.loads(response.content)['results']
```
rlucioni: Same comment about `response.data`.

programs/apps/api/v1/tests/test_views.py
```diff
@@ -149,3 +306,122 @@ def test_create_duplicated_name(self):
...
+        Ensure the API supports listing of Organizations by system users and admins.
+        """
+        org = OrganizationFactory.create()
+        for _ in range(3):
+            CourseCodeFactory.create(organization=org)
+        client = APIClient()
+        client.force_authenticate(user=self._make_user(Role.ADMINS))
+        response = client.get(reverse("api:v1:course-codes-list"))
+        self.assertEqual(response.status_code, 200)
+        results = json.loads(response.content)['results']
```
rlucioni: Same comment about `response.data`.

programs/apps/api/v1/tests/test_views.py
```diff
@@ -149,3 +306,122 @@ def test_create_duplicated_name(self):
...
+
+
+class OrganizationViewTests(JwtMixin, TestCase):
+    """
+    Tests for listing / creating Organizations.
+    """
+
+    def _make_user(self, role_name):
+        """
+        Factory wrapper.
```
rlucioni: The code for this wrapper is repeated below. Can you break it out so the two test cases can share it?

programs/apps/api/v1/views.py
```diff
@@ -36,12 +36,76 @@ class ProgramsViewSet(mixins.CreateModelMixin, mixins.ListModelMixin, viewsets.G
...
+        filters.ProgramStatusQueryFilterBackend,
+        filters.ProgramOrgKeyFilterBackend,
+    )
+    serializer_class = serializers.ProgramSerializer
+
+
+class CourseCodesViewSet(mixins.ListModelMixin, viewsets.GenericViewSet):
+    """
+
+    **Use Cases**
```
rlucioni: What docstring format is this? It looks different than the Napoleon style I've seen us use elsewhere.

programs/apps/programs/models.py
```diff
@@ -54,3 +56,132 @@ class Program(TimeStampedModel):
...
+    class Meta(object):  # pylint: disable=missing-docstring
+        unique_together = ("program", "course_code", "position")
+        ordering = ['position']
+
+    def save(self, *a, **kw):
+        """
+        Override save() to validate m2m cardinality and automatically set the position for a new row.
+        """
+        if not self.id:  # pylint: disable=no-member
+            # before creating, check course code is not associated with another program
```
rlucioni: An organization can't offer two programs which contain the same course code?

programs/apps/programs/models.py
```diff
@@ -54,3 +56,132 @@ class Program(TimeStampedModel):
...
+            if ProgramCourseCode.objects.filter(course_code=self.course_code).exists():
+                raise ValidationError(_('Cannot associate multiple programs with a course code.'))
+            # before creating, ensure that the program has an association with the same org as this course code
+            if not ProgramOrganization.objects.filter(
+                    program=self.program, organization=self.course_code.organization).exists():
+                raise ValidationError(_('Course code must be offered by the same organization offering the program.'))
+            # automatically set position attribute for a new row
+            res = ProgramCourseCode.objects.aggregate(max_position=models.Max('position'))
+            self.position = (res['max_position'] or 0) + 1
```
rlucioni: What's going on here?

programs/apps/programs/models.py
```diff
@@ -54,3 +56,132 @@ class Program(TimeStampedModel):
...
+        help_text=_("The mode_slug value which uniquely identifies the mode in the target LMS."),
+        max_length=64,
+    )
+    # TODO having NULLable sku here keeps things simple, so platform instances
+    # w/o ecommerce setups can reference runs/modes that do not have skus set.
+    # However, we can't enforce a UNIQUE constraint involving sku, because
+    # MySQL does not enforce uniqueness for NULL values.
+    # (see: http://stackoverflow.com/questions/25844786/)
+    # For now no constraint is declared; we may need application-level validation to
+    # properly handle all scenarios.
```
rlucioni: What about overriding `save()` to enforce uniqueness if a SKU is provided?

programs/apps/programs/tests/factories.py
```diff
@@ -3,15 +3,46 @@
...
-        model = Program
 
-    name = factory.Sequence(lambda n: 'test-program-{}'.format(n))  # pylint: disable=unnecessary-lambda
-    description = "test-description"
+class OrganizationFactory(factory.django.DjangoModelFactory):
+    class Meta(object):
+        model = models.Organization
+
+    id = factory.Sequence(lambda n: n)
+    key = factory.LazyAttribute(lambda o: hex(o.id)[::-1].rstrip('0'))
```
rlucioni: Is there a need to convert the ID to a hex string, reverse it, and strip the 0? Why not just use a UUID?

programs/apps/programs/models.py
```diff
@@ -54,3 +56,132 @@ class Program(TimeStampedModel):
...
+
+class CourseCode(TimeStampedModel):
+    """
+    Represents a course independent of run / mode.  This is used to link
+    multiple runs / modes of the same course offering within a program,
+    both for purposes of presentation and in order to enforce logic
+    associated with program completion (i.e. one completed run for each
+    course code in the program indicates completion of the program).
+    """
+    organization = models.ForeignKey(Organization)
```
rlucioni: Some courses available on edx.org are offered jointly by two organizations (e.g., "Visualizing Japan," a collaboration between Harvard and MIT). Should this model be flexible enough to support these kinds of courses?

clintonb:
The status column for programs needs an index.

programs/apps/api/pagination.py
clintonb: Why is this being copied from edx-platform? What is the need for a custom paginator?

programs/apps/api/v1/tests/test_views.py
```diff
@@ -149,3 +306,122 @@ def test_create_duplicated_name(self):
...
+        """
+        org = OrganizationFactory.create()
+        for _ in range(3):
+            CourseCodeFactory.create(organization=org)
+        client = APIClient()
+        client.force_authenticate(user=self._make_user(Role.ADMINS))
+        response = client.get(reverse("api:v1:course-codes-list"))
+        self.assertEqual(response.status_code, 200)
+        results = json.loads(response.content)['results']
+        self.assertEqual(len(results), 3)
```
clintonb: Check the actual values, not just the count.

programs/apps/api/v1/urls.py
```diff
@@ -5,5 +5,30 @@
 
 
 urlpatterns = [
-    url(r'^programs/', views.ProgramsViewSet.as_view({'get': 'list', 'post': 'create'}), name='programs-list'),
+
+    # programs viewset
+    url(
+        r'^programs/$',
+        views.ProgramsViewSet.as_view({'get': 'list', 'post': 'create'}),
```
clintonb: Use [routers](http://www.django-rest-framework.org/api-guide/routers/).

programs/apps/programs/models.py
```diff
@@ -27,8 +28,9 @@ class Program(TimeStampedModel):
         unique=True,
     )
 
-    description = models.TextField(
-        help_text=_('A full-length description of the Program.'),
+    subtitle = models.CharField(
+        help_text=_('A brief, descriptive subtitle for the Program.'),
```
clintonb: sub-title

programs/apps/programs/models.py
clintonb: Those courses have a custom org (e.g. VJx).

programs/apps/programs/models.py
rlucioni: http://www.merriam-webster.com/dictionary/subtitle

programs/apps/programs/models.py
rlucioni: Right, I'd forgotten about that. Would allowing this model to support multiple orgs be premature?

programs/apps/api/pagination.py
jimabramson: All (or at least most) or the HTTP APIs exposed in edx-platform use this pagination style.  I brought it over here in order to maintain consistency.

programs/apps/api/v1/tests/test_views.py
jimabramson: After exploring options I'm inclined to leave this as is.

* Python's `datetime.isoformat()` method produces an output format that cannot actually be replicated using `strftime`.  The `isoformat` method renders timezone offset differently - with a ':' separator - than strftime does.
* DRF's default iso-8601 datetime serializer does `isoformat` under the hood, but, for reasons unknown, strips the offset when it contains all zero values (which it always does since UTC).

So there's no clean way to make DRF output datetimes equal to the way `isoformat` does natively.

programs/apps/programs/models.py
jimabramson: For now, no.  The issue is student dashboard presentation; we are only going to support one program per displayed course.  If we have multiple programs to a course, we won't know which one to display.

programs/apps/api/v1/views.py
jimabramson: https://openedx.atlassian.net/wiki/display/AC/edX+REST+API+Conventions#edXRESTAPIConventions-docstringsDocstringFormat

rlucioni: Thanks.

programs/apps/programs/models.py
jimabramson: it's ensuring the row gets assigned an automatically-incrementing `position` that's unique across all rows with the same program.  Looking at it now, it's got a bug - it should be filtering on the program id.   Will fix.

programs/apps/programs/tests/factories.py
jimabramson: it's a bit of a geekout, but this function creates values that kinda look like edx orgs.  For example: '0x, Fx, BEx'...

programs/apps/programs/models.py
jimabramson: my guess is yes, but I will check w/product.

programs/apps/programs/tests/factories.py
rlucioni: I get it! It's cool, but I'd suggest replacing it with something easier to read; I think it adds unnecessary complexity. That said, if you're intent on keeping it, I'd at least add a comment about what's going on so that others aren't confused about this in the future.

jimabramson: Agree, the latest fixup replaces this with something much less goofy.

programs/apps/programs/models.py
marcotuts: It is fair to presume for now that we map courses to single organizations.

jimabramson: thanks @marcotuts

jimabramson:
@clintonb @rlucioni I think I have all your feedback addressed at this point.

programs/apps/programs/models.py
```diff
@@ -54,3 +56,154 @@ class Program(TimeStampedModel):
...
+
+        TODO (jsa): I'm now convinced that multiple table inheritance is better
+        - one table without SKUs and one table with NOT NULL / UNIQUE SKUs.
+        This compromise should suffice initially.
+        """
+        if self.sku is None and ProgramCourseRunMode.objects.filter(
+                program_course_code=self.program_course_code,
+                course_key=self.course_key,
+                mode_slug=self.mode_slug,
+                sku=None,
```
rlucioni: Should this filter instead check:
```python
ProgramCourseRunMode.objects.filter(
    program_course_code=self.program_course_code,
    course_key=self.course_key,
    mode_slug=self.mode_slug,
    sku=self.sku
).exists()
```
This would also enforce SKU uniqueness.

programs/apps/programs/models.py
```diff
@@ -54,3 +56,154 @@ class Program(TimeStampedModel):
         # default value.  This is consistent with how None/null is handled.
         blank=True,
     )
+
+    class Meta(object):  # pylint: disable=missing-docstring
+        index_together = (('status', 'category'))
```
rlucioni: Nitpick: `index_together` can be a single iterable when dealing with a single set of fields.

rlucioni:
@jimabramson just one question and a nitpick from me.

programs/apps/programs/models.py
jimabramson: If sku isn't None, the database will enforce the unique constraint, so I wouldn't want to do a needless query in that case.

programs/apps/programs/models.py
jimabramson: fixed.

programs/apps/programs/models.py
rlucioni: Cool, I somehow missed the `unique_together` property above. (Which, by the way, can also be a single iterable, since you're only dealing with a single set of fields.)

rlucioni:
:+1:

jimabramson:
@rlucioni @clintonb sorry, just added one last fixup to modify the CourseCode model.

The change is to fully normalize Organizations and CourseCodes by their component key parts.  Before, CourseCodes had a 'key' field with something like 'edX/Demo_Course' while related to the Org with key 'edX'.  Now the CourseCode's key will only contain the 'Demo_Course' part.  This will make for cleaner data and also make it easier to deal with old slash-separated course keys and v1 opaque keys interchangeably (since there won't be any assumptions stored about the key format).

rlucioni:
@jimabramson I agree with the change, but I think the help text for the `key` field could be improved. It currently reads:
```python
'The string prefix of opaque keys for runs of this course in the LMS.'
```
Making this help text clearer and possibly including an example would make this more useful.

jimabramson:
@rlucioni good call, thanks.   I've amended the previous fixup.

rlucioni:
:+1:

clintonb:
1. Rebase on master to get Codecov support.
2. I haven't looked at every single line—we should make smaller PRs—but, what I have seen looks good. Consider that a :+1: from me.

jimabramson:
squashed and rebased.   @clintonb thanks - agree, this PR got too big.  will avoid this going forward...

PR #10, opened 2015-10-14
Added support for static assets

clintonb:
ECOM-2558


clintonb:
FYI @jimabramson @rlucioni

rlucioni:
FYI @AlasdairSwan

PR #12, opened 2015-10-15
Squashed migrations

clintonb:
Manually "squashed" migrations by deleting existing migrations and re-running the command to migration generation command. This is only acceptable since we have not yet deployed the application. Also, field with character length of 256 was update to a length of 255.

ECOM-2186

@jimabramson @rlucioni 

FYI @jibsheet we will need to regenerate the base image for devstack and sandboxes.


rlucioni:
:+1:

jimabramson:
@clintonb looks good.  As long as you're cleaning stuff up - this stuff isn't needed any longer
https://github.com/edx/programs/blob/squash-migrations/programs/apps/programs/migrations/0002_create_groups.py#L18-L27
because we've ditched django model permissions.  We only need to create the groups.  Would you be willing to remove that as well?

jimabramson:
@clintonb also, the 2^n -> 2^n-1 change you made on that course_key field, are there any other fields to which it would make sense to apply that?

jimabramson:
@clintonb ignore me on that last thing about field length, i just found [the SO link you posted](http://stackoverflow.com/questions/1814532/1071-specified-key-was-too-long-max-key-length-is-767-bytes/16820166#16820166) explaining why you did that.

jimabramson:
@clintonb feel free to merge as is, the 0002 migration can be cleaned up in another PR.  :+1: 

@jibsheet this may resolve ECOM-2571

clintonb:
@jimabramson migration updated.

clintonb:
@jimabramson I also update two other charfield limits.

jimabramson:
:+1: :+1:

PR #13, opened 2015-10-19
Serve a page which loads the JS API client

rlucioni:
Only for development and testing of the API client. ECOM-2591. @edx/ecommerce 

@AlasdairSwan, FYI. I've left you a placeholder script tag which you can use to load the client, once available.

### DIFF
--- a/programs/urls.py
+++ b/programs/urls.py
@@ -19,16 +19,23 @@ from django.conf import settings
 from django.conf.urls import include, url
 from django.contrib import admin
 from django.core.urlresolvers import reverse_lazy
-from django.views.generic import RedirectView
+from django.views.generic import RedirectView, TemplateView
 
 from programs.apps.core import views as core_views
 
+
 admin.autodiscover()
+
+js_info_dict = {
+    'packages': ('apps.programs',),
+}
 
 # Always login via edX OpenID Connect
 login = RedirectView.as_view(url=reverse_lazy('social:begin', args=['edx-oidc']), permanent=False, query_string=True)
 
 urlpatterns = [
+    url(r'^i18n/', include('django.conf.urls.i18n')),
+    url(r'^jsi18n/$', 'django.views.i18n.javascript_catalog', js_info_dict),
     url(r'^admin/', include(admin.site.urls)),
     url(r'^api/', include('programs.apps.api.urls', namespace='api')),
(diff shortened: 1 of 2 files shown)


programs/urls.py
```diff
@@ -39,6 +46,16 @@
...
 
-if settings.DEBUG and os.environ.get('ENABLE_DJANGO_TOOLBAR', False):  # pragma: no cover
-    import debug_toolbar  # pylint: disable=import-error
-    urlpatterns.append(url(r'^__debug__/', include(debug_toolbar.urls)))
+# View displaying a page which loads the JS API client
+client = TemplateView.as_view(template_name='client.html')
+
+if settings.DEBUG:  # pragma: no cover
+    urlpatterns += [
+        url(r'^client/$', client, name='client'),
```
clintonb: Why is this debug-only?

programs/urls.py
```diff
@@ -19,16 +19,23 @@
...
-from django.views.generic import RedirectView
+from django.views.generic import RedirectView, TemplateView
 
 from programs.apps.core import views as core_views
 
+
 admin.autodiscover()
 
+js_info_dict = {
+    'packages': (),
```
clintonb: Why no packages specified?

programs/urls.py
rlucioni: It doesn't seem like something we want to expose in a production environment.

clintonb: Got it. Please update the PR to reflect that this is for
development/testing purposes.

-- 
Clinton Blackburn

programs/urls.py
rlucioni: It wasn't clear to me which package to provide here; the docs on this aren't great. I've updated this to specify the `'apps'` package; let me know if that's incorrect.

programs/urls.py
rlucioni: Done.

programs/urls.py
rlucioni: Updated to `'apps.programs'`.

clintonb:
:+1:

jimabramson:
@rlucioni to be a little more specific could we update the name from "client" to "admin_app" or something similar.  otherwise :+1:

rlucioni:
@jimabramson updated, I'm using "author" now.

PR #14, opened 2015-10-20
ECOM-2414 added a simple 'app'

AlasdairSwan:
Added template rendering and app is now hooking into the DOM
Now hitting actual API
Note that pagination has not been plugged in yet - I propose making this its own ticket.

### DIFF
--- a/programs/static/js/models/pagination_model.js
+++ b/programs/static/js/models/pagination_model.js
@@ -1,0 +1,42 @@
+define([
+        'backbone',
+        'jquery',
+        'js/collections/programs_collection'
+    ],
+    function( Backbone, $, ProgramsCollection ) {
+        'use strict';
+
+        return Backbone.Model.extend({
+            urlRoot: '/api/v1/programs/',
+
+            initialize: function() {
+                this.setHeaders();
+            },
+
+            getList: function() {
+                $.ajax({
+                    type: 'GET',
+                    url: this.urlRoot,
+                    headers: this.headers,
+                    contentType: 'application/json; charset=utf-8',
+                    context: this,
+                    success: function( data ) {
+                        var programsCollection = new ProgramsCollection();
+                        programsCollection.set(data.results);
+                        data.results = programsCollection;
+
+                        this.set( data );
+                        this.trigger( 'sync', this );
+                    },
+                    error: function( jqXHR ) {
+                        console.log( 'error: ', jqXHR );
+                    }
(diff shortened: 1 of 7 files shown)


clintonb:
If this JS is loaded on a page not hosted on the programs service, will Backbone need to know the correct hostname/path?

programs/templates/author.html
```diff
@@ -1,9 +1,13 @@
 {# Loads the Programs administration app. Only meant for development and testing of the API client. #}
 
 {% extends 'base.html' %}
+{% load compress %}
+{% load staticfiles %}
```
rlucioni: These directives are present in the base template this template extends, so I don't think you need them here as well.

jimabramson:
@clintonb answer to your question is yes - I think the best thing would be to allow overriding urlRoot from the initialize call, or allowing to specify a baseUrl (domain) to prepend to the urlRoot

programs/static/js/models/pagination_model.js
```diff
@@ -0,0 +1,46 @@
+define([
+        'backbone',
+        'jquery',
+        'js/collections/programs_collection'
+    ],
+    function( Backbone, $, ProgramsCollection ) {
+        'use strict';
+
+        return Backbone.Model.extend({
+            urlRoot: '/api/v1/programs/',
```
rlucioni: The app will also need to make requests to other URLs (e.g., `api/v1/organizations/`, `api/v1/course_codes/`). Should the URL root be a property on the model, or something more flexible than including it here?

rlucioni:
I noticed one of the acceptance criteria on ECOM-2414 is "reading and writing from the API." However, in the interest of keeping PRs small, I'm in favor of keeping this PR limited to reading and adding writing separately.

programs/static/js/models/pagination_model.js
AlasdairSwan: I think we should meet to discuss the scope of this ticket. I originally thought it was just to get something that contacts the API and outputs a UI.

rlucioni: @AlasdairSwan that's correct. I'm just pointing this out in case it should influence the way the app is designed early on.

jimabramson:
:+1: to merge this and close the ticket, once commits are squashed and the JWT hack in setHeaders is removed.

rlucioni:
:+1:

PR #15, opened 2015-10-23
Expose all programs models in the Django admin.

rlucioni:
Also fix uniqueness constraint on ProgramCourseCode. This started more restrictive than it is now. I've intentionally exposed a lot in the admin to give us flexibility as we get this service established. ECOM-2626.

@jimabramson @edx/ecommerce


clintonb:
:+1:

jimabramson:
:+1:

PR #16, opened 2015-10-23
ECOM-2209 Added router and program details view

AlasdairSwan:
Added view only program list and program details views
@rlucioni @jimabramson please review
I realized that I haven't added any tests yet, should we wait until the APIs are more complete and add / edit functionality has been added before adding tests?


AlasdairSwan:
I just added the tests because it was the right thing to do!

programs/static/js/apps/program_admin_app.js
```diff
@@ -2,10 +2,65 @@
...
+                 * and data will be reloaded in the new tab/window.
+                 *
+                 * @param {Event} event - Event being handled.
+                 * @returns {boolean} - Indicates if event handling succeeded (always true).
+                 */
+                navigate: function (event) {
+                    var url = $(event.target).attr('href').replace( this.app.root, '' );
+
+                    // Handle the cases where the user wants to open the link in a new tab/window.
+                    if (event.ctrlKey || event.shiftKey || event.metaKey || event.which === 2) {
```
rlucioni: I'm guessing this 2 here is a keycode, but I'm not sure. Would it make sense to assign it to a constant to make this more readable?

programs/static/js/models/course_list_model.js
```diff
@@ -0,0 +1,25 @@
+define([
+        'backbone'
+    ],
+    function( Backbone ) {
+        'use strict';
+
+        return Backbone.Model.extend({
+            defaults: [
```
rlucioni: Why are these defaults necessary?

AlasdairSwan: Because we don't have an API currently that returns a list of courses. This is just stubbing out that functionality

programs/static/js/test/specs/example_spec.js
```diff
@@ -5,7 +5,7 @@ define([
         'use strict';
 
         // This is an example spec. Delete it once you start writing real tests.
-        describe('jQuery version', function () {
+        xdescribe('jQuery version', function () {
```
rlucioni: Since we now have real tests, I'd heed the comment above and delete this spec.

programs/static/js/test/specs/program_list_spec.js
```diff
@@ -0,0 +1,135 @@
+define([
+        'jquery',
+        'js/models/pagination_model',
+        'js/views/program_list_view'
+    ],
+    function( $, ProgramsModel, ProgramListView ) {
+        'use strict';
+
+
+        // This is an example spec. Delete it once you start writing real tests.
```
rlucioni: Is it though?

programs/static/js/test/specs/program_details_spec.js
```diff
@@ -0,0 +1,133 @@
+define([
+        'js/models/program_model',
+        'js/views/program_details_view'
+    ],
+    function( ProgramModel, ProgramDetailsView ) {
+        'use strict';
+
+
+        // This is an example spec. Delete it once you start writing real tests.
```
rlucioni: This comment is incorrect. I see you copying and pasting!

programs/static/js/apps/program_admin_app.js
AlasdairSwan: I actually copied this from the ecommerce repo, turns out event.which is deprecated. I will clean it up

programs/static/js/views/course_run_view.js
```diff
@@ -0,0 +1,50 @@
...
+                    runs = this.model.get('run_modes'),
+                    updatedRuns = _.filter( runs, function( obj ) {
+                        return obj.start_date !== startDate;
+                    });
+
+                this.model.set({
+                    run_modes: updatedRuns
+                });
+
+                // // Delete this view
```
rlucioni: Nitpick: extra pair of slashes here.

programs/static/js/test/specs/example_spec.js
AlasdairSwan: I did locally, just didn't commit the change :)

programs/static/js/routers/program_router.js
```diff
@@ -0,0 +1,45 @@
...
+            routes: {
+                '': 'programList',
+                'program/:id': 'programDetails'
+            },
+
+            initialize: function( options ) {
+                this.model = options.model;
+            },
+
+            // TODO: prevent zombie views
```
rlucioni: Have you addressed this concern by adding a `removeRun` method to the course run view?

programs/static/js/models/course_list_model.js
rlucioni: So these are meant to be runs associated with a specific course code?

AlasdairSwan: Courses associated with an org.

programs/static/js/test/specs/program_details_spec.js
```diff
@@ -0,0 +1,131 @@
...
+            it( 'should remove a course run when the delete button is clicked', function() {
+                var $runs = view.$el.find('.js-course-runs'),
+                    $removeRunBtn = $runs.find('.js-remove-run').first(),
+                    count = programData.course_codes[0].run_modes.length;
+
+                expect( $runs.find('.js-remove-run').length ).toEqual( count );
+                $removeRunBtn.click();
+
+                setTimeout( function() {
+                    expect( $runs.find('.js-remove-run').length ).toEqual( count - 1 );
```
rlucioni: Why does this need to be wrapped in a timeout?

programs/static/js/routers/program_router.js
AlasdairSwan: No, the issue is that the router is just creating new views as opposed to re-rendering current ones or deleting unused ones. I would prefer to wait and address this once more of the functionality is developed and all of the APIs are plugged in because I think it will be clearer how to handle this then. It shouldn't be a lot of work.

programs/static/js/test/specs/program_details_spec.js
AlasdairSwan: We need to give the code time to execute and update the DOM before testing whether it has done so. Some race conditions on TravisCI can cause issues with this sort of thing.

rlucioni: Got it. It might be worth adding your response above as a comment to the code, for others to refer to in the future.

programs/static/templates/course_run.underscore
```diff
@@ -0,0 +1,5 @@
+<fieldset>
+    <h4>Run <%- course_key %></h4>
+    <button class="js-remove-run">delete</button>
+    <p>Start Date: <%- start_date %></p>
```
rlucioni: Instances of the underlying Django model, `ProgramCourseRunMode`, don't currently have a `start_date` attribute, causing this to throw a reference error when attempting to render course runs.

AlasdairSwan: I needed to add this to test other functionality. I have added some test data that does have a start_date attribute so we can test this. Is the error preventing the app from running for you? I wasn't seeing anything locally.

rlucioni: This doesn't prevent the entire app from running, but it does prevent existing course runs from being displayed in the program detail view.

AlasdairSwan: I have added some code to course_run_view.js which ensures this values is always set, with a comment to remove it once the API has been updated.

rlucioni:
:+1: once commits are squashed.

programs/static/js/test/specs/program_details_spec.js
```diff
@@ -0,0 +1,131 @@
...
+                setTimeout( function() {
+                    view.$el.find('.js-course-select').val('002').trigger('change');
+                }, 100);
+
+                setTimeout( function() {
+                    expect( view.$el.find('.course-details').length ).toEqual( 2 );
+                }, 200);
+
+                jasmine.clock().tick(101);
+                jasmine.clock().tick(201);
```
jimabramson: do you think we need the timeouts to be in 100ms increments even when running locally?   i was wondering if it would be useful to a) set that value from a shared variable somewhere; b) let it be adjusted downward so tests can run faster, if the environment can support it.

programs/static/js/test/specs/program_list_spec.js
```diff
@@ -0,0 +1,133 @@
...
+            afterEach( function() {
+                view.undelegateEvents();
+                view.remove();
+
+                jasmine.clock().uninstall();
+            });
+
+            it( 'should exist', function () {
+                expect( view ).toBeDefined();
+            });
```
jimabramson: do these 'should exist' tests catch anything that the rest of the tests would not?

programs/static/js/test/specs/program_list_spec.js
```diff
@@ -0,0 +1,133 @@
...
+
+            it( 'should render a list of 4 programs', function () {
+                var $ul = view.$el.find('.program-list');
+
+                expect( $ul.find('li').length ).toEqual( listData.count );
+            });
+
+            it( 'should hit the API via the pagniation model: error', function() {
```
jimabramson: pagination

programs/static/js/test/specs/program_list_spec.js
AlasdairSwan: If they fail we know that the error is with the view not rendering, the other tests are specific to the view not doing what we expected. So while other tests would catch the view not rendering / being created, these are a quick and obvious way on confirming we are at least successfully creating the view.

programs/static/js/test/specs/program_list_spec.js
```diff
@@ -0,0 +1,133 @@
...
+                // $.ajax should have been called
+                expect($.ajax).toHaveBeenCalled();
+            });
+
+            it( 'should hit the API via the pagniation model: success', function() {
+                spyOn( $, 'ajax' ).and.callFake( function( event ) {
+                    event.success( listData );
+                });
+
+                view.model.getList();
```
jimabramson: I'm slightly confused about the difference between this and the previous test.  I see that the first one expects a different ajax response, but not why only one has a timeout, nor why it calls getList on `model` in one and `view.model` in the other.

programs/static/js/test/specs/program_details_spec.js
AlasdairSwan: I'm not sure how we would be able to determine the environment specific increments.

programs/static/js/test/specs/program_list_spec.js
AlasdairSwan: That would be a mistake on  my behalf :) Both should call view.model.getList()

programs/static/js/views/course_run_view.js
```diff
@@ -0,0 +1,56 @@
...
+                this.$parentEl = options.$parentEl;
+                this.render();
+            },
+
+            render: function() {
+                this.$el.html( this.tpl( this.data ) );
+                this.$parentEl.append( this.$el );
+            },
+
+            removeRun: function() {
```
jimabramson: start date alone doesn't seem like a totally reliable way to find and remove runs.

programs/static/js/views/program_details_view.js
```diff
@@ -0,0 +1,54 @@
...
+                    });
+                }.bind(this) );
+            },
+
+            addCourse: function() {
+                return new CourseView({
+                    collection: this.model.get('course_codes'),
+                    courseList: this.courseList
+                });
+            }
```
jimabramson: i was expecting to see a way to remove courses too, since we did it for course runs.  is that a small enough addition that it would make sense to tack onto this PR?

programs/static/templates/program_details.underscore
```diff
@@ -0,0 +1,4 @@
+<h2><%- name %></h2>
+<p><%- subtitle %></p>
+<div class="js-course-list"></div>
+<button class="js-add-course">+ Add a course</button>
```
jimabramson: newline issue

programs/static/js/views/program_details_view.js
AlasdairSwan: Just remove from the local state, as opposed to with an API call?

programs/static/js/test/specs/program_details_spec.js
jimabramson: specifically i was considering your earlier comment that the timeouts are designed to avoid race conditions in environments like travis CI.  so one basic thing would be to have timeouts set to 100 when running under CI, but much lower otherwise.

i'm not sure what the right pattern is for setting that value in the jasmine tests, though.

programs/static/js/test/specs/program_list_spec.js
jimabramson: understood.

programs/static/js/test/specs/program_list_spec.js
jimabramson: sorry if i'm dense but why does the success test case need a timeout (but not the other)?

AlasdairSwan: I think it doesn't, I was having context issues with the success call so this is some debugging legacy code. I will remove.

programs/static/js/views/program_details_view.js
jimabramson: local state, as you presently have it for the run.  was there a reason it's only done for the run so far, that i might be missing?

AlasdairSwan: It was just so I could test the adding and removal piece. I didn't want to go too far into the editing work so stopped at removing runs. But it shouldn't be too much work to add it for courses too (and it does make sense to I think).

programs/static/js/test/specs/program_details_spec.js
AlasdairSwan: So I swapped out the 100 with a variable and checked the time it takes to run the tests with intervals of 100 ms and 9000 ms. 
100 ms = 1.97s
9000ms = 1.97s
I don't fully understand how that is possible, I am assuming something to do with the jasmine clock. But I think it shows we don't need to be concerned with the clockInterval value for local development.

programs/static/js/test/specs/program_details_spec.js
```diff
@@ -0,0 +1,132 @@
...
+                    modified: '2015-10-20T18:11:46.854735Z',
+                    name: 'test-program-5',
+                    organizations: [{
+                        display_name: 'test-org-display_name',
+                        key: 'test-org-key'
+                    }],
+                    status: 'unpublished',
+                    subtitle: 'test-subtitle'
+                },
+                clockInterval = 100;
```
jimabramson: could you add a comment mentioning what this is for?  or rename it something very obvious like `testTimeoutInterval`

AlasdairSwan:
@jimabramson I have addressed your comments and courses can now be deleted. Please review once again.

jimabramson:
@AlasdairSwan looks great - :+1:    (pls squash commits before merging)

PR #17, opened 2015-10-23
Allow the Programs authoring app to handle its own routing

rlucioni:
@AlasdairSwan @edx/ecommerce


AlasdairSwan:
:+1:  thanks @rlucioni

PR #19, opened 2015-10-26
Add marketing_slug to Program

ahsan-ul-haq:
ECOM-2576

@edx/ecommerce

### DIFF
--- a/programs/apps/api/serializers.py
+++ b/programs/apps/api/serializers.py
@@ -58,7 +58,8 @@ class ProgramSerializer(serializers.ModelSerializer):
     class Meta(object):  # pylint: disable=missing-docstring
         model = models.Program
         fields = (
-            'id', 'name', 'subtitle', 'category', 'status', 'organizations', 'course_codes', 'created', 'modified'
+            'id', 'name', 'subtitle', 'category', 'status', 'marketing_slug', 'organizations', 'course_codes',
+            'created', 'modified'
         )
         read_only_fields = ('id', 'created', 'modified')
 
--- a/programs/apps/api/v1/tests/test_views.py
+++ b/programs/apps/api/v1/tests/test_views.py
@@ -151,6 +151,7 @@ class ProgramsViewTests(JwtMixin, TestCase):
                 u"id": ANY,
                 u"created": ANY,
                 u"modified": ANY,
+                u'marketing_slug': None,
             }
         )
 
@@ -176,6 +177,7 @@ class ProgramsViewTests(JwtMixin, TestCase):
                     u"id": program.id,
                     u"created": program.created.strftime(DRF_DATE_FORMAT),
                     u"modified": program.modified.strftime(DRF_DATE_FORMAT),
+                    u'marketing_slug': program.marketing_slug,
                 }
             )
 
@@ -202,6 +204,7 @@ class ProgramsViewTests(JwtMixin, TestCase):
                     u"id": program.id,
                     u"created": program.created.strftime(DRF_DATE_FORMAT),
                     u"modified": program.modified.strftime(DRF_DATE_FORMAT),
+                    u'marketing_slug': program.marketing_slug,
                 }
             )
 
@@ -248,6 +251,7 @@ class ProgramsViewTests(JwtMixin, TestCase):
                 u"id": program.id,
                 u"created": program.created.strftime(DRF_DATE_FORMAT),
                 u"modified": program.modified.strftime(DRF_DATE_FORMAT),
+                u'marketing_slug': program.marketing_slug,
             }
         )
 
--- a/programs/apps/programs/models.py
+++ b/programs/apps/programs/models.py
@@ -57,6 +57,28 @@ class Program(TimeStampedModel):
         blank=True,
     )
 
+    marketing_slug = models.CharField(
+        help_text=_('Slug used to generate links to the marketing site'),
+        unique=True,
+        null=True,
+        blank=True,
+        max_length=255
+    )
+
+    def save(self, *a, **kw):
+        """Validate that the marketing slug is not null if program category is
+        xseries and program status is active.
+        """
+        # Checks if the program category is xseries and program status is active.
+        if self.category == constants.ProgramCategory.XSERIES and self.status == constants.ProgramStatus.ACTIVE:
+            # Validate that marketing slug is not null.
+            if not self.marketing_slug:
+                raise ValidationError(
+                    _("Active XSeries Programs must have a valid marketing slug.")
+                )
+
+        return super(Program, self).save(*a, **kw)
+
     class Meta(object):  # pylint: disable=missing-docstring
         index_together = ('status', 'category')
 
--- a/programs/apps/programs/tests/test_models.py
+++ b/programs/apps/programs/tests/test_models.py
@@ -5,7 +5,73 @@ from django.core.exceptions import ValidationError
 from django.test import TestCase
 
 from programs.apps.programs import models
+from programs.apps.programs.constants import ProgramStatus, ProgramCategory
+from programs.apps.programs.models import Program
 from programs.apps.programs.tests import factories
+
+
+class TestProgram(TestCase):
+    """
+    Test for Program model
+    """
+
+    def serialize_program(self, program):
+        """Serialize the the data for the program provided"""
+        return {
+            'name': program.name,
+            'subtitle': program.subtitle,
+            'category': program.category,
+            'status': program.status,
+            'marketing_slug': program.marketing_slug
+        }
+
+    def test_empty_marketing_slug(self):
+        """
+        Ensure that the program could be saved without marketing slug
+        if program category is not xseries or status is UNPUBLISHED. Also
+        verify that marketing slug is required in other case.
+        """
+
+        with self.assertRaises(ValidationError) as context:
+            Program.objects.create(
+                name='test-program',
+                subtitle="test-subtitle",
+                category=ProgramCategory.XSERIES,
+                status=ProgramStatus.ACTIVE,
+            )
+
+        self.assertEqual(
+            "Active XSeries Programs must have a valid marketing slug.",
+            context.exception.message
+        )
+
+        program = {
+            'name': 'test-program',
+            'subtitle': 'test-subtitle',
+            'category': ProgramCategory.XSERIES,
+            'status': ProgramStatus.UNPUBLISHED,
+            'marketing_slug': None
+
+        }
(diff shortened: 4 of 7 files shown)


programs/apps/api/serializers.py
```diff
@@ -73,3 +74,13 @@ def validate_status(self, status):
...
+
+    def validate_marketing_slug(self, marketing_slug):
+        """
+        Prevent Null marketing slug for an ACTIVE program.
+        """
+        if self.validated_data['status'] == constants.ProgramStatus.ACTIVE and not marketing_slug:
+            msg = _("For an active program marketing slug can't be Null.")
+            raise serializers.ValidationError(msg)
+
+        return marketing_slug
```
zubair-arbi: nit: add newline at file end

programs/apps/programs/models.py
```diff
@@ -57,6 +57,26 @@ class Program(TimeStampedModel):
         blank=True,
     )
 
+    marketing_slug = models.CharField(
+        help_text=_('Marketing Slug that can be use to generate links to the marketing site'),
```
zubair-arbi: `'Marketing slug that can be used to generate links for the marketing site'`

programs/apps/programs/models.py
```diff
@@ -57,6 +57,26 @@ class Program(TimeStampedModel):
...
+
+    def save(self, *a, **kw):
+        """Validate that the marketing slug is not null if program category is
+        xseries and program status is active.
+        """
+        # Checks if the program category is xseries and program status is active.
+        if self.category == constants.ProgramCategory.XSERIES and self.status == constants.ProgramStatus.ACTIVE:
+            # Validate that marketing slug is not null.
+            if not self.marketing_slug:
+                raise ValidationError(_("For a publish Program with xseries category Marketing Slug can't be Null."))
```
zubair-arbi: `"Marketing slug 'marketing_slug' must be provided for a published Program with xseries category."`

programs/apps/api/serializers.py
ahsan-ul-haq: @jimabramson I was unable to write test for this validation as when I set status `Active`  (validate_status)[https://github.com/edx/programs/pull/19/files#diff-bdde64b021c3adefae3c2451148f9512R69] is called first and `validate_marketing_slug` is not called. Also as the views doesn't allow to update I could not check this with `PUT` request

jimabramson: @ahsan-ul-haq don't worry about testing validation through the views (POST, PUT etc) - you can just do it in test_models.

zubair-arbi:
lgtm :+1:

programs/apps/programs/models.py
```diff
@@ -57,6 +57,28 @@ class Program(TimeStampedModel):
...
+    def save(self, *a, **kw):
+        """Validate that the marketing slug is not null if program category is
+        xseries and program status is active.
+        """
+        # Checks if the program category is xseries and program status is active.
+        if self.category == constants.ProgramCategory.XSERIES and self.status == constants.ProgramStatus.ACTIVE:
+            # Validate that marketing slug is not null.
+            if not self.marketing_slug:
+                raise ValidationError(
+                    _("Marketing slug 'marketing_slug' must be provided for a published Program with xseries category.")
```
jimabramson: pls remove redundant `'marketing_slug'`

ahsan-ul-haq:
@jimabramson made the suggested change.

programs/apps/api/v1/tests/test_views.py
```diff
@@ -107,7 +113,10 @@ def test_status_list_filter(self, query_status):
         """
         # create one Program of each status
         for status in STATUSES:
-            ProgramFactory(name="{} program".format(status), status=status)
+            if status != ProgramStatus.ACTIVE:
+                ProgramFactory(name="{} program".format(status), status=status)
+            else:
+                ProgramFactory(name="{} program".format(status), status=status, marketing_slug="dummy_slug")
```
jimabramson: 1. why not always add the marketing_slug?  seems like it would be simpler just to always have it, and make it empty for the cases where you need to test that.
2. if the value of marketing_slug really needs to be conditional, please move the logic into the ProgramFactory so it is not repeated multiple times here.

programs/apps/api/v1/tests/test_views.py
```diff
@@ -160,7 +170,10 @@ def test_view_admin(self, status):
         Test that the detail view works correctly for ADMINS, and that deleted
         Programs are filtered out.
         """
-        program = ProgramFactory.create(status=status)
+        if status != ProgramStatus.ACTIVE:
+            program = ProgramFactory.create(status=status)
+        else:
+            program = ProgramFactory.create(status=status, marketing_slug="dummy_slug")
```
jimabramson: see above comment

programs/apps/api/v1/tests/test_views.py
```diff
@@ -186,7 +200,10 @@ def test_view_learner(self, status):
         unpublished and deleted Programs are filtered out.
         """
         filtered_statuses = (ProgramStatus.DELETED, ProgramStatus.UNPUBLISHED)
-        program = ProgramFactory.create(status=status)
+        if status != ProgramStatus.ACTIVE:
+            program = ProgramFactory.create(status=status)
+        else:
+            program = ProgramFactory.create(status=status, marketing_slug="dummy_slug")
```
jimabramson: ditto

programs/apps/programs/tests/test_models.py
```diff
@@ -5,9 +5,36 @@
...
+            factories.ProgramFactory.create(status=ProgramStatus.ACTIVE)
+
+        self.assertEqual(
+            "Marketing slug must be provided for a published Program with xseries category.",
+            context.exception.message
+        )
+
+        factories.ProgramFactory.create(status=ProgramStatus.ACTIVE, marketing_slug="dummy_slug")
+        programs = Program.objects.all()
+        self.assertEqual(len(programs), 1)
```
jimabramson: minor: it would be a stronger test to check that the actual value of marketing_slug in the found result is equal to the expected value.

ahsan-ul-haq:
@jimabramson please take a look I have made the suggested changes.

programs/apps/programs/models.py
```diff
@@ -57,6 +57,28 @@ class Program(TimeStampedModel):
...
+    def save(self, *a, **kw):
+        """Validate that the marketing slug is not null if program category is
+        xseries and program status is active.
+        """
+        # Checks if the program category is xseries and program status is active.
+        if self.category == constants.ProgramCategory.XSERIES and self.status == constants.ProgramStatus.ACTIVE:
+            # Validate that marketing slug is not null.
+            if not self.marketing_slug:
+                raise ValidationError(
+                    _("Marketing slug must be provided for a published Program with xseries category.")
```
rlucioni: `_("Active XSeries Programs must have a valid marketing slug.")`

programs/apps/programs/models.py
```diff
@@ -57,6 +57,28 @@ class Program(TimeStampedModel):
         blank=True,
     )
 
+    marketing_slug = models.CharField(
+        help_text=_('Marketing slug that can be used to generate links for the marketing site'),
```
rlucioni: `help_text=_('Slug used to generate links to the marketing site'),`

programs/apps/programs/tests/test_models.py
```diff
@@ -2,12 +2,83 @@
...
+            category=ProgramCategory.XSERIES,
+            status=ProgramStatus.ACTIVE,
+            marketing_slug='test-slug'
+        )
+
+        programs = Program.objects.all()
+        self.assertEqual(len(programs), 2)
+        self.assertEqual(programs[1].name, 'test-program-2')
+        self.assertEqual(programs[1].status, ProgramStatus.ACTIVE)
+        self.assertEqual(programs[1].marketing_slug, 'test-slug')
```
rlucioni: Instead of making repeated calls to `assertEqual` throughout here, you should be able to populate a dict with the parameters necessary to create the test program, pass the unpacked dict when creating the program, then compare a serialized version of the created program with your original dict.

programs/apps/programs/tests/test_models.py
```diff
@@ -2,12 +2,83 @@
...
+
+    def test_unique_marketing_slug(self):
```
jimabramson: @ahsan-ul-haq this test is causing trouble in Travis but I don't think we really need it.  It's really just testing django's ORM, and they have their own tests :-)

ahsan-ul-haq:
@jimabramson @rlucioni please review.

jimabramson:
@ahsan-ul-haq LGTM :+1:  (once quality issues fixed in the travis build)

programs/apps/programs/tests/test_models.py
```diff
@@ -2,12 +2,88 @@
...
+        self.assertEqual(len(programs), 1)
+        program = {
+            'name': 'test-program',
+            'subtitle': 'test-subtitle',
+            'category': ProgramCategory.XSERIES,
+            'status': ProgramStatus.UNPUBLISHED,
+            'marketing_slug': None
+
+        }
+        self.assertDictEqual(self.serialize_program(programs.first()), program)
```
rlucioni: You can just call `assertEqual` here. Type-specific methods are automatically used by `assertEqual`.

programs/apps/programs/tests/test_models.py
```diff
@@ -2,12 +2,88 @@
...
+        Program.objects.create(
+            name='test-program',
+            subtitle="test-subtitle",
+            category=ProgramCategory.XSERIES,
+            status=ProgramStatus.UNPUBLISHED,
+        )
+
+        programs = Program.objects.all()
+        self.assertEqual(len(programs), 1)
+        program = {
```
rlucioni: To cut down on duplicated lines, declare this dict before creating your test program, then call `Program.objects.create(**programs)` to unpack the dict's contents as keyword arguments to the method.

programs/apps/programs/tests/test_models.py
```diff
@@ -2,12 +2,88 @@
...
+        self.assertDictEqual(self.serialize_program(programs.first()), program)
+
+        Program.objects.create(
+            name='test-program-2',
+            subtitle="test-subtitle-2",
+            category=ProgramCategory.XSERIES,
+            status=ProgramStatus.ACTIVE,
+            marketing_slug='test-slug'
+        )
+        program = {
```
rlucioni: You can use the same dict unpacking approach here.

PR #20, opened 2015-10-26
Provide sample data via a fixture

rlucioni:
This fixture can be used to pre-populate the database with sample data when setting up the service for development or load testing. The fixture sets up two organizations with a program apiece. Each program contains two course codes, and each course code contains two runs. ECOM-2712.

@jimabramson @edx/ecommerce

### DIFF
--- a/docs/getting_started.rst
+++ b/docs/getting_started.rst
@@ -71,6 +71,18 @@ your settings and created the database (if necessary). Migrations can be run wit
 
     $ python manage.py migrate
 
+The Programs repository provides a fixture containing sample data. This fixture can be used to pre-populate the database with sample data when setting up the service for development or load testing. The fixture sets up two organizations with a program apiece. Each program contains two course codes, and each course code contains two runs. If you want to use it, load the data from the fixture as follows:
+
+.. code-block:: bash
+
+    $ python manage.py loaddata sample_data
+
+If you change the Programs schema, please update the fixture. You can do so by installing the fixture on a clean database, applying your new migrations, updating the data as necessary, then running the following command to overwrite the fixture:
+
+.. code-block:: bash
+
+    $ python manage.py dumpdata programs --indent 4 > programs/apps/programs/fixtures/sample_data.json
+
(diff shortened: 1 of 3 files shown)


cpennington:
Would it be better to just set up the load-tests with a startup-step that creates the relevant data for the test?

rlucioni:
@cpennington I think this is more broadly useful.

It would have been helpful to @AlasdairSwan when he was getting started with the implementation of the JS client. It would be useful to new devs who need some sample data to get them familiar with the service. It's also nice to be able to drop your database and reload it with test data without having to manually recreate it each time.

cpennington:
Makes sense.

Any way we can make sure that it doesn't go stale? Do we need a unit tests that just tries to import it, to see if it fails?

jimabramson:
@rlucioni for when the schema changes, is there a defined workflow for how to update the fixtures?  Also does 'loaddata' delete/drop and reload when the data is already present?

rlucioni:
@jimabramson `loaddata` will only overwrite data if the rows created by the fixture already exist. Adding new rows and re-running `loaddata` again won't delete the extra data.

When the schema changes, the easiest way to update the fixture might be to recreate it by installing the fixture on a clean database, applying any new migrations, updating the data as necessary, then running `dumpdata`. I'll add a note about this to the docs.

@cpennington I'll look into adding a unit test to ensure that the fixture stays fresh.

rlucioni:
@cpennington unit test added.

jimabramson:
@rlucioni looks good, with a follow-on request: could one of the programs reference the edX demo course that installs standard with devstack?  I think this will smooth the path for the development use case, since it should light up on the student dashboard right away.

cpennington:
:+1:, although I agree w/ Jim that referencing the demo course would be a nice touch.

jimabramson:
@rlucioni since the PRs that would actually make the demo course light up have not yet merged, I'm OK with merging this as is and updating the data later if you feel better going that route - your call.

:+1:

rlucioni:
@jimabramson @cpennington I've updated the fixture to include a new organization (edX) and a new program (Demo Program) containing one run of the edX demo course (edX/DemoX/Demo_Course).

PR #22, opened 2015-10-28
Update JWT settings when running in a production environment

rlucioni:
Without this, the base JWT settings are used and the configured credentials are never pulled in. ECOM-2627.

@jimabramson @feanil

### DIFF
--- a/programs/settings/production.py
+++ b/programs/settings/production.py
@@ -37,3 +37,9 @@ DB_OVERRIDES = dict(
 
 for override, value in DB_OVERRIDES.iteritems():
     DATABASES['default'][override] = value
+
+JWT_AUTH.update({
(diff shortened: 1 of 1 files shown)


jimabramson:
:+1: 

seems like it might be a good thing to do away with the JWT_ settings, as there's a lot of redundancy with the social auth stuff and easy to miss things like this.

rlucioni:
@jimabramson which JWT_ setting(s) are you referring to? The JWT_AUTH dictionary is required by DRF-JWT. Besides that, we don't have any other JWT_ variables in settings.

jimabramson:
that is the one i'm talking about.   maybe we can find a way to update JWT_AUTH as needed without having to remember to monkey with it in each settings file (a startup module or signal somewhere)?.  would just like to find a way to make this saner.

rlucioni:
@jimabramson does Django support modification of settings at runtime?

PR #23, opened 2015-10-30
Alasdair/ecom 2710 create new program form

AlasdairSwan:
PR is not ready to merge yet, I just want to make sure the tests pass.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -15,7 +15,8 @@ var gulp = require('gulp'),
             'build.js',
             'gulpfile.js',
             'programs/static/js/**/*.js',
-            'programs/static/js/test/**/*.js'
+            'programs/static/js/test/**/*.js',
+            '!programs/bower_components/**/*.js'
         ],
         karamaConf: 'karma.conf.js'
     };
--- a/programs/static/js/models/pagination_model.js
+++ b/programs/static/js/models/pagination_model.js
@@ -29,7 +29,7 @@ define([
             },
 
             setData: function( data ) {
-                this.set( data );
+                this.set( data, { silent: true } );
                 this.setResults( data.results );
             },
 
@@ -41,9 +41,7 @@ define([
                 var programsCollection = new ProgramsCollection();
 
                 programsCollection.set( results );
-                this.set({
-                    results: programsCollection
-                });
+                this.set( { results: programsCollection }, { silent: true });
                 this.trigger( 'sync', this );
             }
         });
--- a/programs/static/js/models/program_model.js
+++ b/programs/static/js/models/program_model.js
@@ -1,9 +1,44 @@
 define([
-        'backbone'
+        'backbone',
+        'jquery',
+        'jquery-cookie'
     ],
-    function( Backbone ) {
+    function( Backbone, $ ) {
         'use strict';
 
-        return Backbone.Model.extend({});
+        return Backbone.Model.extend({
+            url: '/api/v1/programs/',
+
+            save: function() {
+                var method = '',
+                    options = _.extend({validate: true, parse: true}, {
+                        type: 'POST',
+                        url: this.url,
+                        // The API requires a CSRF token for all POST requests using session authentication.
+                        headers: {'X-CSRFToken': $.cookie('programs_csrftoken')},
+                        contentType: 'application/json',
+                        context: this,
+                        // NB: setting context fails in tests
+                        success: _.bind( this.saveSuccess, this ),
+                        error: _.bind( this.saveError, this )
+                    });
+
+                /**
+                 * Simplified version of code from the default Backbone save function
+                 * http://backbonejs.org/docs/backbone.html#section-87
+                 */
+                method = this.isNew() ? 'create' : 'update';
+
+                this.sync( method, this, options);
+            },
+
+            saveError: function( jqXHR ) {
+                this.trigger( 'error', jqXHR );
+            },
+
+            saveSuccess: function() {
+                this.trigger( 'sync', this );
+            }
+        });
     }
 );
--- a/programs/static/js/routers/program_router.js
+++ b/programs/static/js/routers/program_router.js
@@ -1,9 +1,10 @@
 define([
         'backbone',
+        'js/views/program_creator_view',
         'js/views/program_details_view',
         'js/views/program_list_view'
     ],
-    function( Backbone, ProgramDetailsView, ProgramListView ) {
+    function( Backbone, ProgramCreatorView, ProgramDetailsView, ProgramListView ) {
         'use strict';
 
         return Backbone.Router.extend({
@@ -11,6 +12,7 @@ define([
 
             routes: {
                 '': 'programList',
+                'program/new': 'programCreator',
                 'program/:id': 'programDetails'
             },
 
@@ -18,17 +20,28 @@ define([
                 this.model = options.model;
             },
 
-            // TODO: prevent zombie views
-            programList: function() {
-                this.programList = new ProgramListView({
-                    model: this.model
+            programCreator: function() {
+                if ( this.programCreatorView ) {
+                    this.programCreatorView.destroy();
+                }
+
+                this.programCreatorView = new ProgramCreatorView({
+                    programsModel: this.model
                 });
+            },
+
+            programList: function() {
+                if ( !this.programListView ) {
+                    this.programListView = new ProgramListView({
+                        model: this.model
+                    });
+                }
             },
 
             programDetails: function( id ) {
                 var programModel = this.model.get('results').get(id);
 
-                this.programDetails = new ProgramDetailsView({
+                this.programDetailsView = new ProgramDetailsView({
                     model: programModel
                 });
             },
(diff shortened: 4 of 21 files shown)


AlasdairSwan:
@rlucioni @jimabramson please review. Note that validation and the marketing slug are going to be handled in separate PRs.

gulpfile.js
```diff
@@ -11,11 +11,13 @@ var gulp = require('gulp'),
             'programs/static/js/test/**/*.js',
             'programs/static/templates/**/*.js'
         ],
+        // Update to ignore bower_components
```
rlucioni: Looks like you did this.

AlasdairSwan: Good point :)

programs/static/js/models/pagination_model.js
```diff
@@ -41,9 +41,7 @@ define([
                 var programsCollection = new ProgramsCollection();
 
                 programsCollection.set( results );
-                this.set({
-                    results: programsCollection
-                });
+                this.set( { results: programsCollection }, { silent: true });
```
rlucioni: Why silence these events?

programs/static/js/models/program_model.js
```diff
@@ -1,9 +1,45 @@
...
         'use strict';
 
-        return Backbone.Model.extend({});
+        return Backbone.Model.extend({
+            url: '/api/v1/programs/',
+
+            save: function() {
+                var method = '',
+                    options = _.extend({validate: true, parse: true}, {
+                        // Always use POST to avoid having to create a parameterized URL
```
rlucioni: What's wrong with a parametrized URL? How does using POST help avoid it?

programs/static/js/models/program_model.js
```diff
@@ -1,9 +1,45 @@
...
+                        type: 'POST',
+                        url: this.url,
+                        // The API requires a CSRF token for all POST requests using session authentication.
+                        headers: {'X-CSRFToken': $.cookie('programs_csrftoken')},
```
rlucioni: Is using this cookie name going to be a problem when going between stage and prod?

programs/static/js/models/program_model.js
AlasdairSwan: I copied this comment from ecommerce, I will remove it because it's causing too much controversy.

programs/static/js/models/program_model.js
AlasdairSwan: Probably, are there already environment variables exposed JavaScript in the IDA that I could use to determine which cookie to use?

programs/static/js/routers/program_router.js
```diff
@@ -18,17 +20,30 @@ define([
...
+
+                this.programCreatorView = new ProgramCreatorView({
+                    programsModel: this.model
                 });
             },
 
+            programList: function() {
+                if ( this.programListView ) {
+                    this.programListView.destroy();
+                } else {
```
rlucioni: Is this correct? If the list view has already been instantiated, this will destroy it but fail to instantiate a new view.

AlasdairSwan: I don't think that's quite what I wanted, but will look into it some more because there's a slight chance that it was.

programs/static/templates/program_creator_form.underscore
```diff
@@ -0,0 +1,36 @@
...
+
+    <div>
+    <label for="program-name">Name</label>
+        <input id="program-name" class="program-name" aria-describedby="program-name-desc" required>
+        <span id="program-name-desc">The public display name of the program.</span>
+    </div>
+
+    </div>
+        <label for="program-subtitle">Subtitle</label>
+        <input id="program-subtitle" class="program-subtitle" maxlength="140" aria-describedby="program-subtitle-desc" required>
```
rlucioni: The underlying Django model allows a maximum of 255 characters for a program subtitle.

programs/static/js/models/program_model.js
rlucioni: @AlasdairSwan this cookie is only readable by the domain that set it. Stage and prod set their own values for this cookie, so you shouldn't have a problem doing this.

programs/static/js/models/pagination_model.js
jimabramson: i'm going to take a guess that it's because we want to fire 'sync' only once after the collection is set, not for each item in the collection.

do i win the prize?

AlasdairSwan: Yes, when silencing it is because I want to wait until the collection is complete before rendering.

programs/static/templates/program_creator_form.underscore
```diff
@@ -0,0 +1,36 @@
...
+          <% _.each( orgs, function( org ) { %>
+            <option value="<%- org.key %>"><%- org.display_name %></option>
+          <% }); %>
+        </select>
+    </div>
+
+    <div>
+    <label for="program-name">Name</label>
+        <input id="program-name" class="program-name" aria-describedby="program-name-desc" required>
+        <span id="program-name-desc">The public display name of the program.</span>
```
jimabramson: this stuff would need translations.   unless there is some way to get them from the server, since we have help texts and translations there as well.   @rlucioni know if we could share that somehow?

programs/static/js/models/pagination_model.js
rlucioni: Got it, thanks @AlasdairSwan. @jimabramson:

.

programs/static/templates/program_creator_form.underscore
rlucioni: @jimabramson the programs app is already configured to expose Django's [JavaScript translation catalog](https://docs.djangoproject.com/en/1.8/topics/i18n/translation/#using-the-javascript-translation-catalog). The base template already pulls in the catalog, so all @AlasdairSwan should have to do is use the included `gettext` method here.

jimabramson: I was also a little concerned about redundant translations (these messages are very similar to the server side ones but not exactly the same, so translators will have double the work i think).   However, there are so very few cases of this, I'm not really that concerned.

programs/static/templates/program_list.underscore
```diff
@@ -8,3 +8,4 @@
             <p><%- program.category %> Program - <%- program.course_codes.length %> Courses - <%- program.id %></p>
     <% }) %>
 </ul>
+<a href="author/program/new">Create new Program</a>
```
rlucioni: I think href should be `program/new` here. Otherwise the browser is redirected to an invalid path, `author/author/program/new`.

rlucioni: Hmm, maybe not. I was seeing some strange behavior locally that was temporarily resolved by making this change.

programs/static/templates/program_creator_form.underscore
AlasdairSwan: Let's address these validation updates in the sub-task ECOM-2711

rlucioni: Sounds good, I'll make the fix on my branch.

programs/static/templates/program_creator_form.underscore
```diff
@@ -0,0 +1,36 @@
...
+    </div>
+
+    </div>
```
rlucioni: This should be an opening tag, not a closing tag!

AlasdairSwan: yes :)

jimabramson:
FYI as of now i don't have any outstanding issues with this PR apart from some things @rlucioni  has already pointed out.

AlasdairSwan:
@rlucioni how is gettext added to RequireJS? Is it a global JS function?

rlucioni:
@AlasdairSwan as we discussed offline, I don't know that it is. However, it looks like you've addressed that!

AlasdairSwan:
@rlucioni @jimabramson I have added the following file ( programs/static/js/shims/gettext.js ) which is a wrapper around window.gettext to enable it to be included with RequireJS to be more explicit which views rely on it and also allow us to remove the shim from the spec-runner. Let me know if you are happy with this approach and I think we should be good to merge the PR.

programs/static/js/routers/program_router.js
rlucioni: @AlasdairSwan so was this correct? It seems wrong.

gulpfile.js
rlucioni: @AlasdairSwan can you remove this?

programs/static/templates/program_list.underscore
```diff
@@ -8,3 +8,4 @@
             <p><%- program.category %> Program - <%- program.course_codes.length %> Courses - <%- program.id %></p>
```
rlucioni: I think these strings need to be marked for i18n.

rlucioni:
@AlasdairSwan I've made some follow-up comments. I'm fine with your `gettext` shim, but it's raised questions for me about how i18n will work when the app is embedded in Studio. JavaScript code doesn’t have access to .po or .mo files containing the translation catalog; they need to be delivered by the server. Is this a problem we've solved before for the Drupal site?

AlasdairSwan:
It is not a problem we have solved on the drupal site. We were intending on using drupal's translation library and I just created a shim so gettext is a wrapper to that. However we haven't plugged it all together yet.

rlucioni:
@AlasdairSwan got it, I think it's a separate question/problem, anyway.

rlucioni:
:+1:

PR #26, opened 2015-11-02
Update programs README

rlucioni:
I noticed that the Programs README was never updated. I've taken the liberty of updating it and uploading the docs to RTD.

@jimabramson @edx/ecommerce

### DIFF
--- a/docs/programs.rst
+++ b/docs/programs.rst
@@ -1,0 +1,4 @@
+Programs
+========
+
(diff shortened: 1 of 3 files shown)


docs/programs.rst
```diff
@@ -0,0 +1,4 @@
+Programs
```
clintonb: Does this warrant its own page? Why not put it as the second paragraph on the index page?

clintonb:
:+1:

docs/programs.rst
rlucioni: My goal here was to provide a page that could be used to learn about the models and get into specifics about configuration of Programs.

PR #27, opened 2015-11-02
Updated login/logout URLs

clintonb:
- Using the default Django URLs for login and logout
- Using the default auth workflow (OIDC) for browseable API authentication

ECOM-2785


clintonb:

rlucioni:
:+1:

jimabramson:
:+1:

PR #28, opened 2015-11-03
Add validation to the Backbone Program model

rlucioni:
This builds on top of @AlasdairSwan's work in https://github.com/edx/programs/pull/23. ECOM-2711.

I still need to add some tests, but I think this is ready for a look. @AlasdairSwan and @jimabramson, please review.


rlucioni:
Superseded by https://github.com/edx/programs/pull/30.

PR #30, opened 2015-11-05
Add validation to the Backbone Program model and creation view

rlucioni:
The ProgramCreatorView won't save its model unless all validation rules are followed. ECOM-2711.

@AlasdairSwan and @jimabramson, could you please review this?

### DIFF
--- a/programs/static/js/models/program_model.js
+++ b/programs/static/js/models/program_model.js
@@ -9,6 +9,38 @@ define([
         return Backbone.Model.extend({
             url: '/api/v1/programs/',
 
+            // Backbone.Validation rules.
+            // See: http://thedersen.com/projects/backbone-validation/#configure-validation-rules-on-the-model.
+            validation: {
+                name: {
+                    required: true,
+                    maxLength: 64
+                },
+                subtitle: {
+                    // The underlying Django model does not require a subtitle.
+                    maxLength: 255
+                },
+                category: {
+                    required: true,
+                    // XSeries is currently the only valid Program type.
+                    oneOf: ['xseries']
+                },
+                organization: 'validateOrganization'
+            },
+
+            validateOrganization: function(orgArray) {
+                // The array passed to this method contains a single object representing
+                // the selected organization; the object contains the organization's key.
+                // In the future, multiple organizations might be associated with a program.
+                var i;
+
+                for (i = 0; i < orgArray.length; i++) {
+                    if ( orgArray[i].key === 'false' ) {
+                        return gettext('Please select a valid organization.');
+                    }
+                }
+            },
+
             save: function() {
                 var method = '',
                     options = _.extend({validate: true, parse: true}, {
--- a/programs/static/js/test/specs/program_creator_spec.js
+++ b/programs/static/js/test/specs/program_creator_spec.js
@@ -88,6 +88,23 @@ define([
                     view.$el.find('#program-name').val( data.name );
                     view.$el.find('#program-subtitle').val( data.subtitle );
                     view.$el.find('#program-org').val( data.organization );
+
+                    if ( data.category ) {
+                        view.$el.find('#program-type').val( data.category );
+                    }
+                },
+                verifyValidation = function ( data, invalidAttr ) {
+                    var $invalidElement;
+
+                    completeForm( data );
+
+                    view.$el.find('.js-create-program').click();
+
+                    $invalidElement = view.$el.find( '[name="' + invalidAttr + '"]' );
+
+                    expect( view.model.save ).not.toHaveBeenCalled();
+                    expect( $invalidElement.hasClass('invalid') ).toBe( true );
+                    expect( $invalidElement.attr('data-error') ).toBeDefined();
                 };
 
             beforeEach( function() {
@@ -155,6 +172,9 @@ define([
                     event.error();
                 });
 
+                // Fill out the form with valid data so that form model validation doesn't
+                // prevent the model from being saved.
+                completeForm( sampleInput );
                 view.$el.find('.js-create-program').click();
 
                 expect( $.ajax ).toHaveBeenCalled();
@@ -162,7 +182,7 @@ define([
                 expect( view.saveError ).toHaveBeenCalled();
             });
 
-            it( 'should set the model with form data when submitted', function() {
+            it( 'should set the model when valid form data is submitted', function() {
                 completeForm( sampleInput );
 
                 spyOn( $, 'ajax' ).and.callFake( function( event ) {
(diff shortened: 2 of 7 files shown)


programs/static/js/models/program_model.js
```diff
@@ -9,6 +9,39 @@ define([
...
+                },
+                category: {
+                    required: true,
+                    // XSeries is currently the only valid Program type.
+                    oneOf: ['xseries']
+                },
+                organization: 'validateOrganization'
+            },
+
+            validateOrganization: function(value) {
```
AlasdairSwan: It would be clearer to name the variable array

programs/static/js/models/program_model.js
```diff
@@ -9,6 +9,39 @@ define([
...
+                // the selected organization; the object contains the organization's key. In the future,
+                // multiple organizations might be associated with a program.
+                var org;
+
+                /*jshint esnext: true */
+                for (org of value) {
+                    if ( org.key === 'false' ) {
+                        return gettext('Please select a valid organization.');
+                    }
+                }
```
AlasdairSwan: I think this could be simplified using http://underscorejs.org/#has (and we could get rid of the jshint exception comment too)

programs/static/js/test/specs/program_creator_spec.js
```diff
@@ -176,6 +198,50 @@ define([
...
+
+                spyOn( view.model, 'save' );
+
+                // No organization selected.
+                invalidInput.organization = 'false';
+                verifyValidation( invalidInput, 'organization' );
+            });
+
             it( 'should abort the view when the cancel button is clicked', function() {
                 completeForm( sampleInput );
```
AlasdairSwan: If you are calling completeForm in the before each it should be removed from all other tests

programs/static/js/test/specs/program_creator_spec.js
```diff
@@ -176,6 +198,50 @@ define([
...
+                verifyValidation( invalidInput, 'subtitle' );
+            });
+
+            it( 'should not set the model when an invalid category is submitted', function() {
+                var invalidInput = $.extend({}, sampleInput);
+
+                spyOn( view.model, 'save' );
+
+                // Category other than 'xseries' selected.
+                invalidInput.category = 'yseries';
```
AlasdairSwan: 'y-the-patriots-are-awesome-series' is also a good example category name :zap:

programs/static/js/test/specs/program_creator_spec.js
```diff
@@ -176,6 +198,50 @@ define([
...
+                verifyValidation( invalidInput, 'category' );
+            });
+
+            it( 'should not set the model when an invalid organization key is submitted', function() {
+                var invalidInput = $.extend({}, sampleInput);
+
+                spyOn( view.model, 'save' );
+
+                // No organization selected.
+                invalidInput.organization = 'false';
```
AlasdairSwan: I think it would be an empty string instead of a false string.

programs/static/js/views/program_creator_view.js
```diff
@@ -75,6 +88,11 @@ define([
 
             goToListView: function() {
                 Backbone.history.navigate( '', { trigger: true } );
+
+                // Unhook validation.
+                // See: http://thedersen.com/projects/backbone-validation/#unbinding.
+                Backbone.Validation.unbind(this);
```
AlasdairSwan: This should be in the destroy function to make sure validation is unbound whenever the view is destroyed.

programs/static/js/models/program_model.js
rlucioni: Updated.

programs/static/js/models/program_model.js
rlucioni: Really? This isn't checking if the object has a given key; it's checking if the value of the provided org key is "false," meaning that the user failed to select an org from the select menu; "false" is the default value provided in the template.

programs/static/js/test/specs/program_creator_spec.js
rlucioni: I'm not calling `completeForm` in the `beforeEach`. The tests I've added use a helper to cut down on code duplication which takes care of calling `completeForm`, but it's not called in the `beforeEach`.

AlasdairSwan: Sorry, my bad. The way the diff. cut out 20 lines it made it look like you had. As you were :)

programs/static/js/models/program_model.js
AlasdairSwan: So you're checking that ``` value[0].key === 'false' ```

rlucioni: Yes. As discussed offline, I'm using a for loop here to future-proof this logic, in case more than one organization is included in the array. The view, API, and underlying Django model are future-proofed in this way, so the validation logic should be, too.

programs/static/js/views/program_creator_view.js
rlucioni: I originally had it in the `destroy` function, but moved it out after running into the following error while running tests:
```
TypeError: model.associatedViews is undefined in /Users/renzolucioni/edx/programs/programs/static/bower_components/backbone.validation/dist/backbone-validation-amd.js (line 347)
```
The error is due to `model.associatedViews` being undefined [here](https://github.com/thedersen/backbone.validation/blob/master/dist/backbone-validation-amd.js#L347). That code checks if `model.associatedViews` is defined, but the code installed by bower on my machine doesn't:
```javascript
if (view && model.associatedViews.length > 1){
```
Both claim to be v0.11.5. Any suggestions on how I can resolve this?

rlucioni: Clarification: I have the latest v0.11.5 installed, which the fix was supposed to be included with: https://github.com/thedersen/backbone.validation/commit/d76a3366f17dbbddb2b5af8e2389e17bbbd56e3f

rlucioni: Resolved, at least for the sort term. The bug fix was applied to code in the package's src/ directory, but never made it to the package's dist/ directory. I've added a comment to the config file explaining this.

programs/static/js/models/program_model.js
rlucioni: I switched to using non-ES6 syntax to support IE.

programs/static/js/views/program_creator_view.js
rlucioni: Clarification: using the fixed code has allowed me to move unbinding to the destroy method.

rlucioni:
@AlasdairSwan I've addressed all of your comments.

AlasdairSwan:
:+1:  ship it
http://33.media.tumblr.com/b2a004bfc1a7d2ea208417ee78d6c693/tumblr_mhok7pu2r91s1no33o1_400.gif

jimabramson:
@rlucioni everything here LGTM, but don't we need to define messages?

rlucioni:
@jimabramson not necessarily, the validation framework automatically generates messages nice messages. For example, when no name is provided, the message reads "Name is required."

jimabramson:
@rlucioni is i18n supported on the default messages?   If so, :+1:

rlucioni:
@jimabramson good catch, I hadn't thought of that. The default messages are not marked for translation. I'll provided explicit messages wrapped with calls to `gettext`.

rlucioni:
@jimabramson I've marked the validation messages for translation.

jimabramson:
:+1:

PR #31, opened 2015-11-05
Update program attributes

ahsan-ul-haq:
ECOM-2752

* Update Program API to accept patch requests.
* Write Program test to check update program attributes using patch.

@awais786 @zubair-arbi @tasawernawaz @afeef please review.

### DIFF
--- a/programs/apps/api/v1/tests/test_views.py
+++ b/programs/apps/api/v1/tests/test_views.py
@@ -51,7 +51,12 @@ class ProgramsViewTests(JwtMixin, TestCase):
             url = reverse("api:v1:programs-detail", kwargs={'pk': program_id})
         else:
             url = reverse("api:v1:programs-list")
-        return getattr(self.client, method)(url, data=data, HTTP_AUTHORIZATION=auth)
+        if method == 'patch':
+            return getattr(self.client, method)(
+                url, data=data, HTTP_AUTHORIZATION=auth, content_type="application/merge-patch+json"
+            )
+        else:
+            return getattr(self.client, method)(url, data=data, HTTP_AUTHORIZATION=auth)
 
     def test_authentication(self):
         """
@@ -154,6 +159,32 @@ class ProgramsViewTests(JwtMixin, TestCase):
                 u'marketing_slug': None,
             }
         )
+
+    @ddt.data(
+        ({"subtitle": "dummy-subtitle"}, "subtitle"),
+        ({"marketing_slug": "dummy-marketing-slug"}, "marketing_slug"),
+        ({"name": "dummy-name"}, "name")
+    )
+    @ddt.unpack
+    def test_patch(self, program_data, field_updated):
+        """
+        Ensure the API supports Patch request for Programs.
+        """
+        program = ProgramFactory.create()
+        json_data = json.dumps(program_data)
+        response = self._make_request(method="patch", program_id=program.id, data=json_data, admin=True)
+        response_data = json.loads(response.content)
+        self.assertEqual(response_data[field_updated], program_data[field_updated])
+
+    def test_patch_non_admin(self):
+        """
+        Ensure the API support PATCH only for admin users.
+        """
+        # Only allow admin to update program
+        program = ProgramFactory.create()
+        data = json.dumps({"name": "dummy-name"})
+        response = self._make_request(method="patch", program_id=program.id, data=data)
+        self.assertEqual(response.status_code, 403)
 
     @ddt.data(*STATUSES)
     def test_view_admin(self, status):
--- a/programs/apps/api/v1/views.py
+++ b/programs/apps/api/v1/views.py
@@ -4,16 +4,17 @@ Programs API views (v1).
 from rest_framework import mixins, viewsets
 
 from programs.apps.programs import models
-from programs.apps.api import filters, permissions, serializers
+from programs.apps.api import filters, permissions, serializers, parsers
 
 
 class ProgramsViewSet(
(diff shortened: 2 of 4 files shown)


programs/apps/api/v1/tests/test_views.py
```diff
@@ -155,6 +160,41 @@ def test_create(self):
             }
         )
 
+    def test_patch(self):
```
awais786: use ddt for different keys.

programs/apps/api/v1/tests/test_views.py
```diff
@@ -155,6 +160,41 @@ def test_create(self):
...
+        """
+        Ensure the API supports Patch request for Programs
+        """
+        program = ProgramFactory.create()
+        data = json.dumps({"subtitle": "dummy-subtitle"})
+        response = self._make_request(method='patch', program_id=program.id, data=data, admin=True)
+        self.assertEqual(response.status_code, 200)
+        data = json.loads(response.content)
+        self.assertEqual(data['subtitle'], "dummy-subtitle")
```
awais786: Also assert the whole response.content with some predefined dict.

programs/apps/api/v1/tests/test_views.py
```diff
@@ -155,6 +160,41 @@ def test_create(self):
             }
         )
 
+    def test_patch(self):
+        """
+        Ensure the API supports Patch request for Programs
+        """
+        program = ProgramFactory.create()
+        data = json.dumps({"subtitle": "dummy-subtitle"})
```
zubair-arbi: nit: Please be consistent with the prefix, either user `dummy-` or `test-` for all strings

programs/apps/api/v1/tests/test_views.py
```diff
@@ -155,6 +160,41 @@ def test_create(self):
...
+        response = self._make_request(method='patch', program_id=program.id, data=data, admin=True)
+        self.assertEqual(response.status_code, 200)
+        data = json.loads(response.content)
```
zubair-arbi: nit: rename `data` -> `response_data`

zubair-arbi:
looks good :+1:

ahsan-ul-haq:
@jimabramson please review.
FYI @rlucioni

programs/apps/api/v1/tests/test_views.py
```diff
@@ -51,7 +51,12 @@ def _make_request(self, method='get', program_id=None, data=None, admin=False):
             url = reverse("api:v1:programs-detail", kwargs={'pk': program_id})
         else:
             url = reverse("api:v1:programs-list")
-        return getattr(self.client, method)(url, data=data, HTTP_AUTHORIZATION=auth)
+        if method == 'patch':
+            return getattr(self.client, method)(
+                url, data=data, HTTP_AUTHORIZATION=auth, content_type="application/json"
```
jimabramson: I think "application/json" is actually correct for all request methods; however, just had a conversation with some folks here about how we do PATCH.   The content_type in that case should be "application/merge-patch+json" for this case.  Take a look at the details here: https://openedx.atlassian.net/wiki/display/AC/edX+REST+API+Conventions#edXRESTAPIConventions-3.HTTPVerbs

In order to use that content type we need to add an extra bit to DRF to understand it.  There is an example in edx-platform [here](https://github.com/edx/edx-platform/blob/master/openedx/core/lib/api/parsers.py#L80-L84).  I suggest this should be added to a new module, programs/apps/api/parsers.py

programs/apps/api/v1/views.py
```diff
@@ -32,6 +33,15 @@ class ProgramsViewSet(
         Only users with global administrative rights may create programs.  POST requests from non-
         admins will result in status 403.
 
+        # Update existing Program.
+        PATCH /api/v1/programs/{program_id}
+
+        If the request is successful, the HTTP status will be 200 and the response body will
+        contain a JSON-formatted representation of the newly-updated program field.
```
jimabramson: minor correction: s/newly-updated program field/newly-updated program/

programs/apps/api/v1/views.py
```diff
@@ -39,6 +49,7 @@ class ProgramsViewSet(
         * subtitle: A brief, descriptive subtitle for the Program.
         * category: The category / type of Program.  Right now the only value allowed is 'xseries'.
         * status: The lifecycle status of this Program.  Right now the only value allowed is 'unpublished'.
+        * marketing_slug: Slug used to generate links to the marketing site.
```
jimabramson: thanks for fixing that

jimabramson:
@ahsan-ul-haq looks solid.  Other than changes for merge-patch (see inline), looks good to go

programs/apps/api/v1/tests/test_views.py
rlucioni: @jimabramson just so I'm clear, we can use DRF's default PATCH handling, but we need to add support for the merge patch media type?

ahsan-ul-haq: @jimabramson I guess that would be reasonable as per the conventions used in the edX. Thanks for sharing the resources.

ahsan-ul-haq:
@jimabramson suggested changes have been addressed. It would be great if you could take a look again.

jimabramson:
:+1: 
@rlucioni to your [question](https://github.com/edx/programs/pull/31#discussion_r44079944) - yes, correct.
@AlasdairSwan FYI - client API calls should use content type "application/merge-patch+json" when updating individual fields, based on the discussion/changes in this PR.

PR #35, opened 2015-11-09
Make JWT authentication handler more robust

rlucioni:
Django's get_or_create is vulnerable to a race condition which can cause user creation to fail. Retrying when failures occur should cause the object responsible for the initial failure to be retrieved. ECOM-2840.

@jimabramson please review.


jimabramson:
:+1:

PR #36, opened 2015-11-09
Log details when 401 responses are returned.

jimabramson:
The hope is that this will shed light on the cause of sporadic 401 errors.

ECOM-2839

@rlucioni  please review.

### DIFF
--- a/programs/apps/api/exception_handler.py
+++ b/programs/apps/api/exception_handler.py
@@ -1,0 +1,22 @@
+"""
+Custom exception handling for API Views.
+"""
+import logging
+
+from rest_framework.exceptions import APIException
+from rest_framework.status import HTTP_401_UNAUTHORIZED
+from rest_framework.views import exception_handler
+
+
+logger = logging.getLogger(__name__)
+
+
+def auth_exception_handler(exc, context):
+    """
+    Wrap default exception handling with diagnostic logging for auth failures.
+    """
+    if isinstance(exc, APIException) and exc.status_code == HTTP_401_UNAUTHORIZED:
+        exc_type = u'{}.{}'.format(exc.__class__.__module__, exc.__class__.__name__)
+        logger.info("raised_http_401: exc_type=%r, exc_detail=%r", exc_type, exc.detail)
+
+    return exception_handler(exc, context)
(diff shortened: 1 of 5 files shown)


programs/apps/api/exception_handler.py
```diff
@@ -0,0 +1,22 @@
+"""
+Custom exception handling for API Views.
+"""
+
+from rest_framework.exceptions import APIException
+from rest_framework.status import HTTP_401_UNAUTHORIZED
+from rest_framework.views import exception_handler
+
+import logging
```
rlucioni: Nitpick: standard library import, should be above the DRF imports.

programs/apps/api/tests/test_exception_handler.py
```diff
@@ -0,0 +1,86 @@
...
+Tests for custom exception handling in the API.
+"""
+
+from django.conf.urls import url
+from django.test import override_settings, TestCase
+import mock
+from rest_framework import exceptions, status
+from rest_framework.test import APIClient
+from rest_framework.views import APIView
+from testfixtures import LogCapture
```
rlucioni: I think `testfixtures` needs to be added to the test requirements.

programs/apps/api/tests/test_exception_handler.py
```diff
@@ -0,0 +1,86 @@
...
+
+    @mock.patch('rest_framework.views.APIView.initial', side_effect=Custom401Exception("oops"))
+    def test_custom_401(self, _mock_get_response):
+        """
+        Ensure we emit a log message when exceptions with a 401 status code get
+        raised.
+        """
+        self._check_response(401, 'programs.apps.api.tests.test_exception_handler.Custom401Exception', 'oops')
+
+    @override_settings(ROOT_URLCONF='programs.apps.api.tests.test_exception_handler')
```
rlucioni: Is this necessary? You're already setting `ROOT_URLCONF` on the class.

rlucioni:
@jimabramson would it be cleaner to do something like the following, in `authentication.py`?
```python
class JwtAuthentication(JSONWebTokenAuthentication):
    def authenticate(self, request):
        try:
            return super(JwtAuthentication, self).authenticate(request)
        except Exception as e:
            logger.info(e)
            raise
...
```

jimabramson:
@rlucioni the reason I did not go that route was out of concern that the exception could be getting raised somewhere else.  One conceivable cause of 401s is not finding the JWT auth header at all, for example.

jimabramson:
rebased

rlucioni:
@jimabramson understood. :+1:

PR #37, opened 2015-11-10
Alasdair/ecom 2770 adding style to the app

AlasdairSwan:
@rlucioni @jimabramson this PR pulls in the Pattern Library, scopes it to within our app and adds style to what we have so far.
@talbs here is the PL being pulled into an IDA PR!

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -1,16 +1,14 @@
 'use strict';
 
-var gulp = require('gulp'),
+var del = require('del'),
+    gulp = require('gulp'),
     jscs = require('gulp-jscs'),
     jshint = require('gulp-jshint'),
     KarmaServer = require('karma').Server,
+    sass = require('gulp-sass'),
     path = require('path'),
     paths = {
-        spec: [
-            'programs/static/js/**/*.js',
-            'programs/static/js/test/**/*.js',
-            'programs/static/templates/**/*.js'
-        ],
+        karmaConf: 'karma.conf.js',
         lint: [
             'build.js',
             'gulpfile.js',
@@ -18,7 +16,15 @@ var gulp = require('gulp'),
             'programs/static/js/test/**/*.js',
             '!programs/bower_components/**/*.js'
         ],
-        karamaConf: 'karma.conf.js'
+        spec: [
+            'programs/static/js/**/*.js',
+            'programs/static/js/test/**/*.js',
+            'programs/static/templates/**/*.js'
+        ],
+        styles: {
+            src: 'programs/static/sass/**/*.scss',
+            dest: 'programs/static/css/'
+        }
     };
 
 /**
@@ -53,6 +59,19 @@ gulp.task('jscs', function () {
 });
 
 /**
+ * Compiles CSS from src Sass files
+ * first it removes the output folder to start clean
+ */
+gulp.task('css', function() {
+    del([ paths.styles.dest ])
+        .then( function() {
+            return gulp.src( paths.styles.src )
+                .pipe( sass().on( 'error', sass.logError ) )
+                .pipe( gulp.dest( paths.styles.dest ) );
+        });
+});
+
+/**
  * Monitors the source and test files, running tests
  * and linters when changes detected.
  */
@@ -60,4 +79,4 @@ gulp.task('watch', function () {
     gulp.watch(paths.spec, ['test', 'lint', 'jscs']);
 });
 
-gulp.task('default', ['test']);
+gulp.task('default', ['test', 'css']);
--- a/programs/static/js/shims/edx-icons.js
+++ b/programs/static/js/shims/edx-icons.js
@@ -1,0 +1,16 @@
+define([
+        'afontgarde'
+    ],
+    function( AFontGarde ) {
+        'use strict';
+
+        /**
+         * Adds the 'edx-icons' class to the DOM's HTML element
+         * if the icon fonts successfully load
+         */
+        return new AFontGarde('edx-icons', {
+            // Check a few random icons to see if our font loads
+            glyphs: '\uE621\uE622\uE623'
+        });
+    }
+);
--- a/programs/static/sass/_views.scss
+++ b/programs/static/sass/_views.scss
@@ -1,0 +1,61 @@
+
+// ------------------------------
+// Programs: Views
+
+// About: styling for specific views.
+
+// ------------------------------
+// #PROGRAM LISTS
+// ------------------------------
+.program-list {
+    list-style-type: none;
+    padding-left: 0;
+
+    .program-details {
+        .name {
+            font-size: 2rem;
+        }
+
+        .status {
+            @include float(right);
+        }
+
+        .category {
+            color: palette(grayscale, base);
+        }
+    }
+}
+
+.app-header {
+    border-bottom: 1px solid palette(grayscale, base);
+    margin-bottom: 20px;
+}
+
+.course-container {
+    .subtitle {
+        color: palette(grayscale, base);
+    }
+}
+
+.run-container {
+    position: relative;
+    margin: {
+        bottom: 20px;
+    };
+
+    &:before {
+        content: '';
+        width: 5px;
+        height: calc( 100% + 1px );
+        background: palette(grayscale, base);
+        position: absolute;
+        top: 0;
+        left: 0;
+    }
+}
+
+.course-container {
+    margin: {
+        bottom: 20px;
+    };
+}
--- a/programs/static/sass/main.scss
+++ b/programs/static/sass/main.scss
@@ -1,1 +1,0 @@
-// Styles for the programs service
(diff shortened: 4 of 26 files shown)


programs/static/js/shims/edx-icons.js
```diff
@@ -0,0 +1,12 @@
+define([
+        'afontgarde'
+    ],
+    function() {
```
rlucioni: I think you need to pass `AFontGarde` in here.

AlasdairSwan: I also thought that, but it wasn't doing what I expected. I will take another look.

programs/static/sass/main.scss
```diff
@@ -1 +1,41 @@
...
+@import 'utilities';
+*/
+.#{$scope-class} {
+    @import 'components';
+    @import 'layouts';
+    @import 'views';
+}
+/*
+@import 'print';
+@import 'overrides';
```
rlucioni: Should these and other commented out imports be removed?

programs/static/templates/course_details.underscore
```diff
@@ -1,11 +1,16 @@
...
+<div class="card course-container">
 <% if ( display_name ) {  %>
-    <h3><%- display_name %></h3>
-    <button class="js-remove-course"><%= gettext('delete') %></button>
-    <p><%- organization.display_name %> / <%- key %>
+    <span class="copy copy-large emphasized"><%- display_name %></span>
+    <button class="js-remove-course btn delete right">
+        <span class="icon-fallback icon-fallback-text">
+            <span class="icon icon-trash-o" aria-hidden="true"></span>
+            <span class="text">Remove <%- display_name %> from the program</span>
```
rlucioni: This text should be marked for translation.

programs/static/templates/course_run.underscore
```diff
@@ -1,5 +1,10 @@
-<fieldset>
-    <h4>Run <%- course_key %></h4>
-    <button class="js-remove-run">delete</button>
-    <p>Start Date: <%- start_date %></p>
-</fieldset>
+<div class="card run-container">
+    <span class="copy copy-large emphasized">Run <%- course_key %></span>
```
rlucioni: This text should be marked for translation.

programs/static/templates/course_run.underscore
```diff
@@ -1,5 +1,10 @@
...
+    <button class="js-remove-run btn delete right">
+        <span class="icon-fallback icon-fallback-text">
+            <span class="icon icon-trash-o" aria-hidden="true"></span>
+            <span class="text">Remove run <%- course_key %> from the program</span>
```
rlucioni: This text should be marked for translation.

programs/static/templates/course_run.underscore
```diff
@@ -1,5 +1,10 @@
...
+        </span>
+    </button>
+    <p class="copy copy-base subtitle">Start Date: <%- start_date %></p>
```
rlucioni: This text should be marked for translation.

rlucioni:
@AlasdairSwan can we take this opportunity to add styles for the `invalid` class appended by Backbone.Validation?

programs/static/sass/main.scss
AlasdairSwan: Not at this stage, I would like to keep them until just before launch of the admin UI in case we need to add these partials so we are following the same standard as the Pattern Library.

AlasdairSwan:
@rlucioni can we add that as a separate sub-task and PR? In the interests of keeping PRs small, and also to get the base styles into the repo sooner rather than later.

programs/static/sass/main.scss
rlucioni: Understood. Could you please add a comment to that effect, explaining why these are being kept commented?

programs/static/js/shims/edx-icons.js
AlasdairSwan: But now I see that failing test I will need to resolve the issue :)

rlucioni:
@AlasdairSwan sure, I've filed https://openedx.atlassian.net/browse/ECOM-2859.

gulpfile.js
```diff
@@ -1,16 +1,14 @@
...
     KarmaServer = require('karma').Server,
+    sass = require('gulp-sass'),
     path = require('path'),
     paths = {
-        spec: [
-            'programs/static/js/**/*.js',
-            'programs/static/js/test/**/*.js',
-            'programs/static/templates/**/*.js'
-        ],
+        karamaConf: 'karma.conf.js',
```
jimabramson: assume that should be `karmaConf`

AlasdairSwan: Good catch, yes it should. Looks like it was karamaConf earlier so weird that it ever worked! I will fix.

programs/static/js/shims/edx-icons.js
```diff
@@ -0,0 +1,12 @@
+define([
+        'afontgarde'
+    ],
+    function( AFontGarde ) {
+        'use strict';
+
+        AFontGarde('edx-icons', { // jshint ignore:line
+            // Check a few random icons to see if our font loads
+            glyphs: '\uE621\uE622\uE623'
+        });
```
jimabramson: @AlasdairSwan could you explain what's going on in here?

AlasdairSwan: This is taken from the Pattern Library codebase. AFontGarde is an external library which trues to load icon fonts, on success it adds the class 'edx-icons' to the DOM HTML element. I will add a comment explaining this.

programs/static/sass/_views.scss
```diff
@@ -0,0 +1,61 @@
+
+// ------------------------------
+// Programs: Layouts
+
+// About: styling for generic layouts and responsive support.
```
jimabramson: copied comment needs update

programs/static/sass/main.scss
```diff
@@ -1 +1,42 @@
...
+// #CONFIG:         app-centric configuration/overrides
+// #LIB:            third party libraries and dependencies
+// #EXTENSIONS:     add-ons and further customizations
+
+
+// ------------------------------
+// #CONFIG
+// ------------------------------
+$scope-class: 'program-app';
+$layout-direction: ltr;
```
jimabramson: just for my own edification, when / how might this be overridden to rtl ?

AlasdairSwan: Good questions. In LMS and studio we currently support this, so there are 2 output css files based on this. This will add an additional layer of complexity that I didn't notice, in that studio will need to load either the ltr or rtl version of the programs CSS.

AlasdairSwan: And I will need to update this PR to output both versions of CSS.

programs/static/templates/program_creator_form.underscore
```diff
@@ -1,44 +1,50 @@
...
-        <label for="program-marketing-slug"><%= gettext('Marketing Slug') %></label>
-        <input id="program-marketing-slug" class="program-marketing-slug" name="marketing_slug" aria-describedby="program-marketing-slug-desc">
-        <span id="program-marketing-slug-desc">
-            <%= gettext('Slug used to generate links to the marketing site.') %>
-        </span>
-    </div>
+        <div class="field">
+            <label class="field-label" for="program-marketing-slug"><%= gettext('Marketing Slug') %></label>
+            <input id="program-marketing-slug" class="field-input input-text program-marketing-slug" name="marketing_slug" aria-describedby="program-marketing-slug-desc">
+            <div class="field-hint" id="rogram-marketing-slug-desc">
```
jimabramson: typo in id

programs/static/sass/main.scss
jimabramson: ok, glad i asked :-)

rlucioni:
:+1:

programs/static/templates/program_details.underscore
```diff
@@ -1,4 +1,12 @@
-<h2><%- name %></h2>
-<p><%- subtitle %></p>
-<div class="js-course-list"></div>
-<button class="js-add-course">+ <%= gettext('Add a course') %></button>
+<header class="app-header">
+    <h2><%- name %></h2>
+    <p><%- subtitle %></p>
```
jimabramson: as discussed over IM there may be a problem with the font being used for this element.

AlasdairSwan:
@jimabramson I have updated so there are 2 CSS files being compiled (main-ltr and main-rtl) but I am not sure how we would determine which to use. Has i18n been taken care of for the app yet? I am assuming the work would fall under that? (i.e. is it enough for this PR to have the 2 CSS files exist and handling loading of the correct version in a future i18n story?)

programs/static/templates/program_details.underscore
AlasdairSwan: I have fixed this in the app for now (issues with a scoped reset partial) but I may need to revisit once we load the app in studio. It is expected that there will be a bit of clean-up required anyway once we see it in studio.

jimabramson:
@AlasdairSwan it makes sense to me that we'll handle loading of the correct version when we plug this into studio.

jimabramson:
:+1:

PR #38, opened 2015-11-11
Save ProgramOrganization with Program in creation endpoint

zubair-arbi:
ECOM-2727
@ahsan-ul-haq @awais786 @afeef @tasawernawaz 

Save Program with provided valid `Organization` (with its `key`) or raise validation errors.
* POST request for endpoint `/programs/` expects a nested array containing exactly one object in the `'organizations'` field, and it contains a 'key' which references an org known to the system.
* POST request for endpoint `/programs/` returns a 400, if:
 * the `key` does not match the key of an organization in the db
 * the length of the array `'organizations'` is anything other than 1

### DIFF
--- a/programs/apps/api/serializers.py
+++ b/programs/apps/api/serializers.py
@@ -21,6 +21,17 @@ class OrganizationSerializer(serializers.ModelSerializer):
         fields = ('display_name', 'key')
 
 
+class ProgramOrganizationSerializer(serializers.ModelSerializer):
+    """Serializer for the organizations for the Program Serializer."""
+    display_name = serializers.CharField(source='organization.display_name', read_only=True)
+    key = serializers.CharField(source='organization.key')
+
+    class Meta(object):  # pylint: disable=missing-docstring
+        model = models.ProgramOrganization
+        fields = ('display_name', 'key')
+        read_only_fields = ('display_name', )
+
+
 class ProgramCourseRunModeSerializer(serializers.ModelSerializer):
     """Serializer for the program course run mode model."""
 
@@ -63,14 +74,52 @@ class ProgramSerializer(serializers.ModelSerializer):
         )
         read_only_fields = ('id', 'created', 'modified')
 
-    organizations = OrganizationSerializer(many=True, read_only=True)
+    organizations = ProgramOrganizationSerializer(many=True, source='programorganization_set')
     course_codes = ProgramCourseCodeSerializer(many=True, read_only=True, source='programcoursecode_set')
+
+    def create(self, validated_data):
+        """
+        Create a Program and link it with the provided organization.
+        """
+        programs_organizations_data = validated_data.pop('programorganization_set')
+        program = super(ProgramSerializer, self).create(validated_data)
+
+        # get organizations with the provided parameter 'key' and attach it
+        # with the newly created program
+        for organization_data in programs_organizations_data:
+            org_data = organization_data.get('organization')
+            organization = models.Organization.objects.get(key=org_data.get('key'))
+            models.ProgramOrganization.objects.get_or_create(program=program, organization=organization)
+
+        return program
 
     def validate_status(self, status):
         """
         Prevent creation of Programs with a status other than UNPUBLISHED.
         """
         if self.instance is None and status != constants.ProgramStatus.UNPUBLISHED:
-            msg = _("When creating a Program, \"{status}\" is not a valid choice.")
-            raise serializers.ValidationError(msg.format(status=status))
+            error_msg = _("When creating a Program, '{status}' is not a valid choice.")
+            raise serializers.ValidationError(error_msg.format(status=status))
+
         return status
+
+    def validate_organizations(self, organizations):
+        """
+        Prevent creation of Programs without a valid organization.
+        """
+        error_msg = _("Provide exactly one valid/existing Organization while creating a Program.")
+
+        # validate that user has provided only one organization
+        # Note: This is a temporary, application-level constraint for single organization.
+        if self.instance is None and len(organizations) != 1:
+            raise serializers.ValidationError(error_msg)
+
+        # validate that the provided organization exists in database with
+        # the same key
+        for organization_data in organizations:
+            org_data = organization_data.get('organization')
+            if not models.Organization.objects.filter(key=org_data.get('key')).exists():
+                error_msg = _("Provided Organization with key '{org_key}' doesn't exist.")
+                raise serializers.ValidationError(error_msg.format(org_key=org_data.get('key')))
+
+        return organizations
--- a/programs/apps/api/v1/tests/test_views.py
+++ b/programs/apps/api/v1/tests/test_views.py
@@ -54,12 +54,33 @@ class ProgramsViewTests(JwtMixin, TestCase):
             url = reverse("api:v1:programs-detail", kwargs={'pk': program_id})
         else:
             url = reverse("api:v1:programs-list")
+
+        content_type = "application/json"
         if method == 'patch':
-            return getattr(self.client, method)(
-                url, data=data, HTTP_AUTHORIZATION=auth, content_type="application/merge-patch+json"
-            )
-        else:
-            return getattr(self.client, method)(url, data=data, HTTP_AUTHORIZATION=auth)
+            data = json.dumps(data)
+            content_type = "application/merge-patch+json"
+        elif method == 'post':
+            data = json.dumps(data)
+
+        return getattr(self.client, method)(
+            url, data=data, HTTP_AUTHORIZATION=auth, content_type=content_type
+        )
+
+    def _validate_org_data_errors(self, request_data, validation_error, organizations_data=None):
+        """
+        DRY helper for validation of error responses while creating programs
+        with invalid organizations data.
+        """
+        if organizations_data is not None:
+            request_data["organizations"] = organizations_data
+
+        response = self._make_request(method='post', data=request_data, admin=True)
+        self.assertEqual(response.status_code, 400)
+        response_data = json.loads(response.content)
+        self.assertEqual(
+            response_data['organizations'],
+            [validation_error]
+        )
 
     def test_authentication(self):
         """
@@ -142,10 +163,16 @@ class ProgramsViewTests(JwtMixin, TestCase):
 
     def test_create(self):
         """
-        Ensure the API supports creation of Programs.
+        Ensure the API supports creation of Programs with a valid organization.
         """
+        # Create a valid organization
+        OrganizationFactory.create(key="test-org-key", display_name="test-org-display_name")
+
         data = self._build_post_data()
+        # Add the valid organization in POST data while creating a Program
+        data["organizations"] = [{"key": "test-org-key"}]
         response = self._make_request(method='post', data=data, admin=True)
+
         self.assertEqual(response.status_code, 201)
         self.assertEqual(
             json.loads(response.content),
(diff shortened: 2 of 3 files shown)


programs/apps/api/serializers.py
```diff
@@ -21,6 +21,16 @@ class Meta(object):  # pylint: disable=missing-docstring
         fields = ('display_name', 'key')
 
 
+class ProgramOrganizationSerializer(serializers.ModelSerializer):
+    """Serializer for the organizations for the Program Serializer."""
+    key = serializers.CharField(max_length=64)
+
+    class Meta(object):  # pylint: disable=missing-docstring
+        model = models.Organization
+        fields = ('display_name', 'key')
```
ahsan-ul-haq: As `Organization Model` already have `key` field then why are you creating `CharField` again?

programs/apps/api/serializers.py
```diff
@@ -63,14 +73,48 @@ class Meta(object):  # pylint: disable=missing-docstring
...
+        """
+        # validate that user has provided only one organization
+        if self.instance is None and len(organizations) != 1:
+            msg = _("Provide only one valid/existing Organization while creating a Program.")
+            raise serializers.ValidationError(msg)
+
+        # validate that the provided organization exists in database with
+        # the same key
+        org_key = organizations[0].get('key')
+        if models.Organization.objects.filter(key=org_key).count() != 1:
```
ahsan-ul-haq: Use `models.Organization.objects.filter(key=org_key).exists()`.
For reference https://docs.djangoproject.com/en/1.8/ref/models/querysets/#django.db.models.query.QuerySet.exists

programs/apps/api/v1/views.py
```diff
@@ -50,6 +51,8 @@ class ProgramsViewSet(
         * category: The category / type of Program.  Right now the only value allowed is 'xseries'.
         * status: The lifecycle status of this Program.  Right now the only value allowed is 'unpublished'.
         * marketing_slug: Slug used to generate links to the marketing site.
+        * organizations: List data containing organizations with which the program is associated.
+        * course_codes: List data containing course codes with which the program is associated.
```
ahsan-ul-haq: Nit: Please remove this it should only contain values that could be add/update.

programs/apps/api/serializers.py
zubair-arbi: @ahsan-ul-haq since there is constraint `unique=True` on `key` field of Organization so thats why I have to add this field in serializer. without this I get following error:
```python
{u'key': [u'This field must be unique.']}
```

programs/apps/api/v1/views.py
zubair-arbi: ok

programs/apps/api/serializers.py
zubair-arbi: @ahsan-ul-haq this is due to avoid exceptions:
`DoesNotExist: Organization matching query does not exist.`

zubair-arbi:
@rlucioni @jimabramson Please review.

programs/apps/api/serializers.py
zubair-arbi: Sorry about my above comment. I will make the requested change.

ahsan-ul-haq:
LGTM :+1:

programs/apps/api/serializers.py
```diff
@@ -21,6 +21,16 @@ class Meta(object):  # pylint: disable=missing-docstring
         fields = ('display_name', 'key')
 
 
+class ProgramOrganizationSerializer(serializers.ModelSerializer):
+    """Serializer for the organizations for the Program Serializer."""
+    key = serializers.CharField(max_length=64)
+
+    class Meta(object):  # pylint: disable=missing-docstring
+        model = models.Organization
```
jimabramson: This approach seems confusing - I would expect a serializer with this name to have `ProgramOrganization` as its model.  In fact I think that should be possible - the `ProgramCourseCodeSerializer` follows this pattern, and its fields point to the related model (`CourseCode`) using the `source` parameter.  Worth a try?

programs/apps/api/serializers.py
jimabramson: That error suggests that `save()` is trying to create a new Organization instead of locating and linking an existing one.  If so, that won't be an issue if `ProgramOrganizationSerializer` uses `ProgramOrganization` as its model, because it will just be creating a link to the existing Org identified by the `key`.

programs/apps/api/serializers.py
```diff
@@ -63,14 +73,48 @@ class Meta(object):  # pylint: disable=missing-docstring
...
+        Create a Program and link it with the provided organization.
+        """
+        organizations_data = validated_data.pop('organizations')
+        program = super(ProgramSerializer, self).create(validated_data)
+
+        for org_data in organizations_data:
+            # get organization with the provide 'key' and attach it with the
+            # newly created program
+            organization = models.Organization.objects.get(key=org_data.get('key'))
+            models.ProgramOrganization.objects.get_or_create(program=program, organization=organization)
```
jimabramson: I suspect that if the `ProgramOrganizationSerializer` can be changed according to my suggestion above, this custom behavior will no longer be needed.

programs/apps/api/serializers.py
```diff
@@ -63,14 +73,48 @@ class Meta(object):  # pylint: disable=missing-docstring
...
+
         return status
+
+    def validate_organizations(self, organizations):
+        """
+        Prevent creation of Programs without a valid organization.
+        """
+        # validate that user has provided only one organization
+        if self.instance is None and len(organizations) != 1:
+            msg = _("Provide only one valid/existing Organization while creating a Program.")
```
jimabramson: nitpick: s/only/exactly/

programs/apps/api/v1/tests/test_views.py
```diff
@@ -333,7 +405,7 @@ def test_missing_fields(self, field):
         else:
             self.assertIn("field is required", content[field][0])
 
-    @ddt.data(ProgramStatus.ACTIVE, ProgramStatus.RETIRED, ProgramStatus.DELETED, None, "", " ", "unrecognized")
+    @ddt.data(ProgramStatus.ACTIVE, ProgramStatus.RETIRED, ProgramStatus.DELETED, "", " ", "unrecognized")
```
jimabramson: why was `None` removed in this test (and the following one)?

jimabramson:
@zubair-arbi views and tests LGTM; had some questions about the serializer approach.

programs/apps/api/serializers.py
zubair-arbi: @jimabramson Sure. I will try this approach.

programs/apps/api/v1/tests/test_views.py
zubair-arbi: Previously for tests we were not json encoding POST data and also not setting any `content_type` https://github.com/edx/programs/blob/master/programs/apps/api/v1/tests/test_views.py#L62.
So while adding list data for `organizations` the django's default test client was changing this list value from
```python
'organizations': [{'key': 'test-org-key'}]
```
 to a unicode string
```
python 'organizations': [u"{'key': 'test-org-key'}"]
```
which is the reason why I was always getting empty list for organizations in tests.

So after json dumping POST data and setting `content_type = "application/json"` two tests starts failing for `None` value:
```python
======================================================================
FAIL: test_create_with_invalid_status_4_None (programs.apps.api.v1.tests.test_views.ProgramsViewTests)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/zubairafzal1/Workspace/devstack/programs_env/lib/python2.7/site-packages/ddt.py", line 146, in wrapper
    return func(self, *args, **kwargs)
  File "/Users/zubairafzal1/Workspace/devstack/programs/programs/apps/api/v1/tests/test_views.py", line 417, in test_create_with_invalid_status
    self.assertIn("not a valid choice", content["status"][0])
AssertionError: 'not a valid choice' not found in u'This field may not be null.'
```
Since this ddt value was getting different response than others so it made sense to remove it. If you think we still need to check this value then I can add a separate test.

programs/apps/api/serializers.py
zubair-arbi: @jimabramson we still need this manual validation since we are displaying `ProgramOrganization` object with custom fields `display_name`, `key` and also adding it with custom way by providing organizations key's list only `[{"key": "org-key"}]`.

zubair-arbi:
@jimabramson Please review. I have tried you suggested approach but we will still need custom create for `ProgramOrganization` since the provided post data for `organizations` lets say
```python
'organizations: [{"key": "test-org-key"}]
```
will be converted to:
```python
u'programorganization_set': [OrderedDict([(u'organization', {u'key': u'TEST'})])]
```

programs/apps/api/serializers.py
```diff
@@ -21,6 +21,17 @@ class Meta(object):  # pylint: disable=missing-docstring
...
 
+class ProgramOrganizationSerializer(serializers.ModelSerializer):
+    """Serializer for the organizations for the Program Serializer."""
+    display_name = serializers.CharField(source='organization.display_name', read_only=True)
+    key = serializers.CharField(source='organization.key')
+
+    class Meta(object):  # pylint: disable=missing-docstring
+        model = models.ProgramOrganization
+        fields = ('display_name', 'key')
+        read_only_fields = ('display_name', )
```
jimabramson: minor: declaring `fields` and `read_only_fields` in Meta seems redundant with the explicit field declarations above.  Does everything work the same if they are removed?

jimabramson:
@zubair-arbi you're right about custom `create()` and `update()` for nested writeable representations, my mistake.  (DRF docs: http://www.django-rest-framework.org/api-guide/serializers/#writable-nested-representations)

programs/apps/api/serializers.py
zubair-arbi: @jimabramson if I remove `fields` and `read_only_fields` then organizations data is displayed as below:
```python
"organizations": [
    {
        "id": 1,
        "display_name": "Testing organization",
        "key": "TEST",
        "created": "2015-11-06T10:45:14.321921Z",
        "modified": "2015-11-06T10:45:14.329739Z",
        "program": 3,
        "organization": 1
    }
],
```

jimabramson: sorry, my mistake again :-) thanks for giving it a try

zubair-arbi:
@cpennington please give a look. thanks

programs/apps/api/serializers.py
zubair-arbi: No problem. :)

programs/apps/api/v1/tests/test_views.py
jimabramson: understood, thanks for the explanation

jimabramson:
@zubair-arbi create implementation looks good to me.  Where does this leave behavior for PATCH and PUT requests?  Since we have no custom `update` in the serializer, it's unclear how that will be handled.  (I'm OK with implementing PATCH/PUT support for the orgs in a separate ticket, but I think we need comments or tests that show how the API will behave until then.)

programs/apps/api/serializers.py
```diff
@@ -63,14 +74,50 @@ class Meta(object):  # pylint: disable=missing-docstring
...
+        """
+        error_msg = _("Provide exactly one valid/existing Organization while creating a Program.")
+
+        # validate that user has provided only one organization
+        if self.instance is None and len(organizations) != 1:
+            raise serializers.ValidationError(error_msg)
+
+        # validate that the provided organization exists in database with
+        # the same key
+        organization = organizations[0].get('organization')
```
rlucioni: Since you've already validated that `len(organizations) != 1` by this point, I'd use a for loop here to iterate across `organizations`. This will make it easier to remove the single-organization constraint in the future.

programs/apps/api/serializers.py
```diff
@@ -63,14 +74,50 @@ class Meta(object):  # pylint: disable=missing-docstring
...
+
         return status
+
+    def validate_organizations(self, organizations):
+        """
+        Prevent creation of Programs without a valid organization.
+        """
+        error_msg = _("Provide exactly one valid/existing Organization while creating a Program.")
+
+        # validate that user has provided only one organization
```
rlucioni: Please note that this is a a temporary, application-level constraint.

programs/apps/api/v1/views.py
```diff
@@ -62,7 +64,7 @@ class ProgramsViewSet(
         filters.ProgramOrgKeyFilterBackend,
     )
     serializer_class = serializers.ProgramSerializer
-    parser_classes = (parsers.MergePatchParser,)
+    parser_classes = (parsers.MergePatchParser, JSONParser)
```
rlucioni: For clarity, I would change the imports above to:
```python
from rest_framework import parsers as drf_parsers

from programs.apps.api import filters, permissions, serializers, parsers as edx_parsers
```
And then here:
```python
parser_classes = (edx_parsers.MergePatchParser, drf_parsers.JSONParser)
```

programs/apps/api/serializers.py
zubair-arbi: added note below

zubair-arbi:
@jimabramson the ticket was related to only creation endpoint `post`. We can add code for `update` and `patch` with a separate ticket. Would you like to add a `update` method with `TODO` as a placeholder until then?

rlucioni:
@zubair-arbi :+1: from me, thanks for making those changes.

jimabramson:
@zubair-arbi never mind, you can ignore me about the update stuff.  :+1:

PR #39, opened 2015-11-12
Alasdair/ecom 2754 editable fields

AlasdairSwan:
@jimabramson @rlucioni please review. There is a failing test that I am debugging but you can start reviewing as is because any changes should be minor (i.e. tests pass locally and functionality works so it's probably an issue with the test that is failing).

### DIFF
--- a/programs/static/js/test/specs/program_details_spec.js
+++ b/programs/static/js/test/specs/program_details_spec.js
@@ -1,8 +1,9 @@
 define([
+        'jquery',
         'js/models/program_model',
         'js/views/program_details_view'
     ],
-    function( ProgramModel, ProgramDetailsView ) {
+    function( $, ProgramModel, ProgramDetailsView ) {
         'use strict';
 
         describe('ProgramDetailsView', function () {
@@ -38,6 +39,7 @@ define([
                     }],
                     created: '2015-10-20T18:11:46.854451Z',
                     id: 5,
+                    marketing_slug: 'test-program-slug',
                     modified: '2015-10-20T18:11:46.854735Z',
                     name: 'test-program-5',
                     organizations: [{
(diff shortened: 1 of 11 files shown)


programs/static/js/test/specs/program_details_spec.js
```diff
@@ -70,76 +132,114 @@ define([
...
+                    jasmine.clock().tick( ( testTimeoutInterval * 2 ) + 1 );
+                });
             });
 
-            it( 'should set course details on change of the course select', function() {
-                var $btn = view.$el.find('.js-add-course').first();
+            describe( 'Edit data', function() {
+                it( 'should enable a user to edit the name, subtitle and marketing slug fields', function() {
+                    editField( '.program-name', 'name' );
+                    //editField( '.program-subtitle', 'subtitle' );
```
rlucioni: Looks like this should be uncommented.

programs/static/js/utils/validation_config.js
```diff
@@ -0,0 +1,68 @@
...
+            number: gettext( '{0} must be a number' ),
+            email: gettext( '{0} must be a valid email' ),
+            url: gettext( '{0} must be a valid url' ),
+            inlinePattern: gettext( '{0} is invalid' )
+        });
+
+        _.extend( Backbone.Validation.callbacks, {
+            // Gets called when a previously invalid field in the
+            // view becomes valid. Removes any error message.
+            // Should be overridden with custom functionality.
```
rlucioni: You can probably get rid of this line.

programs/static/js/utils/validation_config.js
```diff
@@ -0,0 +1,68 @@
...
+                      .removeAttr( 'data-error' );
+
+                $message.removeClass( errorClass )
+                        .find( messageContent )
+                        .html( '' );
+            },
+
+            // Gets called when a field in the view becomes invalid.
+            // Adds a error message.
+            // Should be overridden with custom functionality.
```
rlucioni: Same here, this can go.

rlucioni:
Looks nice @AlasdairSwan, :+1: once tests are passing.

programs/static/js/test/specs/program_details_spec.js
```diff
@@ -47,7 +49,62 @@ define([
...
+                    var $input = view.$el.find( el ),
+                        $btn = view.$el.find( '.js-add-course' ),
+                        title = $input.val();
+
+                    editField( el, update );
+
+                    $btn.focus();
+                    $input.blur();
+
+                    expect( title ).not.toEqual( update );
```
jimabramson: confused about what's being tested here - we captured the value of `$input.val` before updating, but we didn't actually set it to anything.  it seems then this only confirms that `testUpdatedFieldBlur()` was called with a nonempty `update`?

programs/static/js/views/program_details_view.js
```diff
@@ -1,14 +1,16 @@
...
         'js/models/pagination_model',
         'js/views/course_details_view',
         'text!templates/program_details.underscore',
-        'gettext'
+        'gettext',
+        'js/utils/validation_config'
     ],
-    function( Backbone, $, _, CourseListModel, ProgramsModel,
+    function( Backbone, BackboneValidation, $, _, CourseListModel, ProgramsModel,
               CourseView, ListTpl ) {
```
jimabramson: gettext and validation_config aren't added to the signature, is that intentional?

jimabramson:
I have a question about how a test is working, otherwise LGTM

programs/static/js/views/program_details_view.js
AlasdairSwan: Yes, with methods that are added globally they don't need to be.

AlasdairSwan:
@jimabramson thanks for catching that test weirdness, I have updated to test the input value before and after the field edit to make sure the field value is actually being updated to the new value.

jimabramson:
:+1:

PR #40, opened 2015-11-13
adjust JWT_LEEWAY and use custom JWT decode function

jimabramson:
ECOM-2839

Process of elimination points to the "iat" (issued at) claim in programs JWT tokens as the cause of intermittent bursts of 401 errors.  This is the only claim with a dynamic value present in these tokens that is not present in ecom tokens (which do not experience such errors).  It's a timestamp on the generation of the JWT.   If the clock on a host where the JWT is generated is even a fraction of a second ahead of the clock on the host where it's read, the JWT can appear to have been generated in the future, and thus be considered invalid.

"Leeway" is a setting which allows the JWT library to tolerate drift between clocks on two systems - this PR changes it from the library default (zero tolerance) to 1 second.

Also in this PR are changes to require the presence of "exp" and "iat" claims, and to get more specific log messages for invalid JWT tokens in case fixing the leeway doesn't completely address the problem.

@rlucioni @cpennington @clintonb

### DIFF
--- a/programs/apps/api/jwt_decode_handler.py
+++ b/programs/apps/api/jwt_decode_handler.py
@@ -1,0 +1,42 @@
+"""
+Custom JWT decoding function for django_rest_framework jwt package.
+
+Adds logging to facilitate debugging of InvalidTokenErrors.  Also
+requires "exp" and "iat" claims to be present - the base package
+doesn't expose settings to enforce this.
+"""
+import logging
+
+import jwt
+from rest_framework_jwt.settings import api_settings
+
+logger = logging.getLogger(__name__)
+
+
+def decode(token):
+    """
+    Ensure InvalidTokenErrors are logged for diagnostic purposes, before
+    failing authentication.
+    """
+
+    options = {
+        'verify_exp': api_settings.JWT_VERIFY_EXPIRATION,
+        'require_exp': True,
+        'require_iat': True,
+    }
+
+    try:
+        return jwt.decode(
+            token,
+            api_settings.JWT_SECRET_KEY,
+            api_settings.JWT_VERIFY,
+            options=options,
+            leeway=api_settings.JWT_LEEWAY,
+            audience=api_settings.JWT_AUDIENCE,
+            issuer=api_settings.JWT_ISSUER,
+            algorithms=[api_settings.JWT_ALGORITHM]
+        )
+    except jwt.InvalidTokenError as exc:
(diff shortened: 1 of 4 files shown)


programs/apps/api/v1/tests/mixins.py
```diff
@@ -27,13 +27,20 @@ def generate_token(self, payload, secret=None):
         token = jwt.encode(payload, secret)
         return token
 
-    def generate_id_token(self, user, admin=False, ttl=0):
+    def generate_id_token(self, user, admin=False, ttl=1, **overrides):
```
jimabramson: note to reviewers: changing default ttl from 0 to 1 addresses an intermittent timing bug that would occasionally cause tests to fail, since expiration was being set to now (`time.time()`).

clintonb:
You'll also need to update edx/configuration.

:+1:

jimabramson:
@clintonb nice catch but I'm not sure it's actually necessary - we don't seem to be overriding anything in `JWT_AUTH` from configuration for programs.  Could I be missing something?

clintonb:
@jimabramson Never mind. I see what you did there: https://github.com/edx/programs/blob/master/programs/settings/production.py#L41-L45.

programs/apps/api/jwt_decode_handler.py
```diff
@@ -0,0 +1,42 @@
...
+            token,
+            api_settings.JWT_SECRET_KEY,
+            api_settings.JWT_VERIFY,
+            options=options,
+            leeway=api_settings.JWT_LEEWAY,
+            audience=api_settings.JWT_AUDIENCE,
+            issuer=api_settings.JWT_ISSUER,
+            algorithms=[api_settings.JWT_ALGORITHM]
+        )
+    except jwt.InvalidTokenError, exc:
```
rlucioni: Could you use the `Exception as variable` syntax here? It's forward compatible with Python 3 and reads a little better, too.

programs/apps/api/v1/tests/mixins.py
rlucioni: @AlasdairSwan this is the cause of that flaky test you saw yesterday. Thanks for fixing this @jimabramson.

rlucioni:
:+1:

programs/apps/api/jwt_decode_handler.py
jimabramson: oops, nice catch.  old habits...

PR #43, opened 2015-11-17
Support updating programs via PATCH requests

rlucioni:
ECOM-2752.

@jimabramson, please review. @AlasdairSwan, FYI. This will require the authoring app to submit data only for fields being modified. For example, if the subtitle is changed from "foo" to "bar," the API will expect a PATCH request with the following payload:
```javascript
{'subtitle': 'bar'}
```

### DIFF
--- a/programs/apps/api/v1/tests/test_views.py
+++ b/programs/apps/api/v1/tests/test_views.py
@@ -14,6 +14,7 @@ from programs.apps.api.v1.tests.mixins import AuthClientMixin, JwtMixin
 from programs.apps.core.constants import Role
 from programs.apps.core.tests.factories import UserFactory
 from programs.apps.programs.constants import ProgramCategory, ProgramStatus
+from programs.apps.programs.models import Program
 from programs.apps.programs.tests.factories import (
     CourseCodeFactory,
     OrganizationFactory,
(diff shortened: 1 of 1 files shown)


AlasdairSwan:
@rlucioni because this PR will break functionality until the front end has been updated would it make sense to hold off on merging it until the PR for the front end update is ready to go too?

rlucioni:
@AlasdairSwan sure, that makes sense. I should note that the PATCH handler added here is capable of accepting request payloads with multiple keys, if multiple fields are to be changed. For example:
```javascript
{
    'name': 'baz',
    'subtitle': 'bar'
}
```
However, since the front end currently includes fields like `organizations` and `course_codes` in its request payloads, things would still break. I'll hold off on merging this until the front end no longer submits those values in PATCH requests.

programs/apps/api/serializers.py
```diff
@@ -93,6 +93,17 @@ def create(self, validated_data):
...
+    def update(self, instance, validated_data):
+        """
+        Update a program.
+        """
+        for field, value in validated_data.viewitems():
+            setattr(instance, field, value)
+
+        instance.save()
+
+        return instance
```
jimabramson: @rlucioni question; if a client submits nested objects as part of the PATCH request will this cover those updates as well?

jimabramson: never mind, you answered my question above.

jimabramson:
:+1:

rlucioni:
Turns out I unwittingly re-implemented DRF's [default update method](https://github.com/tomchristie/django-rest-framework/blob/master/rest_framework/serializers.py#L871-L882). I will scrap these changes but leave the updated, more correct tests in place.

PR #44, opened 2015-11-17
Remove uniqueness constraint on Program marketing_slug field

rlucioni:
Requiring this field to be unique while simulatenously allowing null values in the database means that only one object is allowed to have an empty marketing_slug.

@jimabramson @AlasdairSwan


AlasdairSwan:
Once rebased with master :+1:

jimabramson:
@rlucioni I'm open to this, but it's definitely not true in MySQL than only one row can have a NULL value for a unique-constrained column.  Is it a problem with sqlite perhaps?

rlucioni:
@jimabramson good point, thanks for correcting me. The issue is actually that the authoring app is submitting an empty string when the user leaves the `marketing_slug` field blank, which the server then saves as the value assigned to that field.

I still think there's value to removing this uniqueness constraint. If someone creates a Program, but then decided to (soft-)delete it and create a new one - this is a use-case the Program statuses support - they won't be able to recycle the marketing slug used previously. I'll create a ticket for implementing the correct behavior: sending `null` from the client and setting `None` on the model.

rlucioni:
The same argument makes me wonder if we should also remove the uniqueness constraint on program `name`. Thoughts?

jimabramson:
@rlucioni While I think the unique constraint on the marketing slug is of little value, that column is going to go away in the not-too-distant future.  I'm fine with removing it and avoiding issues we shouldn't be worrying about.

Regarding the name field: if we remove the constraint we expose ourselves to duplicate names unless we implement custom logic to enforce uniqueness of name across all non-deleted programs.  The alternative is we keep the unique constraint in place, and instead, force changing the program name when it is soft-deleted (e.g. append '-deleted' to it).  That will be a less frequent use case, ensures cleaner data, and is easy to work around (e.g. django admin) so that's the route I'd prefer.  Agree?

rlucioni:
@jimabramson that makes sense. In light of our offline conversation, I'm going to go ahead and close this. Teaching the server to handle the empty string correctly is the more correct way of solving this problem.

PR #46, opened 2015-11-17
Updating data for a push

AlasdairSwan:
@rlucioni @jimabramson here is the update to make it so that PATCH only sends the changed data. Thus involved less code than I first thought.

### DIFF
--- a/programs/static/js/models/program_model.js
+++ b/programs/static/js/models/program_model.js
@@ -48,11 +48,10 @@ define([
                 }
             },
 
-            save: function( options ) {
-                var method = '',
-                    patch = options && options.patch ? true : false,
+            getConfig: function( options ) {
+                var patch = options && options.patch ? true : false,
                     params = patch ? this.get('id') : '',
-                    config = _.extend({validate: true, parse: true}, {
+                    config = _.extend({ validate: true, parse: true }, {
                         type: patch ? 'PATCH' : 'POST',
                         url: this.urlRoot + params,
                         // The API requires a CSRF token for all POST requests using session authentication.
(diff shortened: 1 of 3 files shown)


jimabramson:
nice :+1:

programs/static/js/models/program_model.js
```diff
@@ -70,6 +70,10 @@ define([
                  */
                 method = this.isNew() ? 'create' : (patch ? 'patch' : 'update');
 
+                if ( patch ) {
+                    config.data = JSON.stringify( options.update ) || this.attributes;
```
rlucioni: Looks like this is the line causing coverage to fail.

AlasdairSwan: Yeah, I am just pulling out the config into a getConfig function that I can write tests for

AlasdairSwan: And makes the code neater as an added bonus.

programs/static/js/test/specs/program_details_spec.js
```diff
@@ -241,6 +241,27 @@ define([
...
+
+                it( 'should create a POST config object by default', function() {
+                    var config = view.model.getConfig();
+
+                    expect( config.type ).toEqual( 'POST' );
+                    expect( config.contentType ).toEqual( 'application/json' );
+                    expect( config.data ).not.toBeDefined();
+                });
+
+                it( 'should create a PATCH config object by default', function() {
```
rlucioni: I'd strike the "by default" part. It should create a PATCH config object when patching is explicitly requested.

rlucioni:
Just a minor comment about tests, otherwise :+1: .

PR #48, opened 2015-11-19
Alasdair/ecom 2958 publish validation

AlasdairSwan:
@jimabramson @rlucioni please review.

### DIFF
--- a/programs/static/js/views/program_details_view.js
+++ b/programs/static/js/views/program_details_view.js
@@ -92,13 +92,21 @@ define([
             confirmPublish: function( event ) {
                 event.preventDefault();
 
-                this.modalView = new ModalView({
-                    model: this.model,
-                    callback: _.bind( this.publishProgram, this ),
-                    content: this.getModalContent(),
-                    parentEl: '.js-publish-modal',
-                    parentView: this
-                });
+                /**
+                 * Update validation to make marketing slug required
+                 * Note that because this validation is not required for
+                 * the program creation form and is only happening here
+                 * it makes sense to have the validation at the view level
+                 */
+                if ( this.model.isValid( true ) && this.validateMarketingSlug() ) {
+                    this.modalView = new ModalView({
+                        model: this.model,
+                        callback: _.bind( this.publishProgram, this ),
+                        content: this.getModalContent(),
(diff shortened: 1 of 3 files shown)


programs/static/js/views/program_details_view.js
```diff
@@ -145,6 +153,27 @@ define([
...
+                    $message = {};
+
+                if ( this.model.get( 'marketing_slug' ).length > 0 ) {
+                    isValid = true;
+                } else {
+                    $input = this.$el.find( '#program-marketing-slug' );
+                    $message = $input.siblings( '.field-message' );
+
+                    // Update DOM
+                    $input.addClass( 'has-error' );
```
rlucioni: Should we be using `errorClass` from `validation_config.js`?

programs/static/js/views/program_details_view.js
```diff
@@ -145,6 +153,27 @@ define([
...
+                    $message.addClass( 'has-error' );
+                    $message.find( '.field-message-content' )
```
rlucioni: Similarly, should this be `messageContent` from `validation_config.js`?

rlucioni: And on a related note, am I correct in saying that Backbone.Validation is clearing these error classes when validation succeeds?

AlasdairSwan: Yes it is clearing the message. I will check whether messageContent is available here, if so I will use it.

programs/static/js/views/program_details_view.js
AlasdairSwan: These variables are scoped to the validation_config.js so are not available here.

programs/static/js/views/program_details_view.js
AlasdairSwan: The variables are scoped to the validation_config.js so are not available here.

AlasdairSwan: Also, as this field and validation are a short term fixed I'm not too worried about having the error classes defined in multiple places.

rlucioni:
:+1:

jimabramson:
LGTM :+1:

PR #49, opened 2015-11-19
Alasdair/ecom 2766 disable details edit for non unpublished courses

AlasdairSwan:

### DIFF
--- a/programs/static/js/views/program_details_view.js
+++ b/programs/static/js/views/program_details_view.js
@@ -39,14 +39,15 @@ define([
             },
 
             postRender: function() {
-                var courses = this.model.get('course_codes');
+                var courses = this.model.get( 'course_codes' );
 
                 _.each( courses, function( course ) {
-                    var title = course.key + 'Course';
+                    var title = course.key + 'Course',
+                        data = $.extend( course, { programStatus: this.model.get( 'status' ) });
 
                     this[ title ] = new CourseView({
                         collection: courses,
-                        data: course
+                        data: data
                     });
                 }.bind(this) );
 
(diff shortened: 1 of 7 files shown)


programs/static/templates/program_details.underscore
```diff
@@ -32,8 +32,8 @@
             </div>
 
             <div class="js-inline-edit field">
-                <label for="program-subtitle" class="sr-only"><%= gettext('Marketing Slug') %></label>
                 <span class="js-model-value copy copy-base subtitle"><%- marketing_slug %></span>
+                <label for="program-subtitle" class="sr-only"><%= gettext('Marketing Slug') %></label>
```
jimabramson: did the moved lines in this file change behavior, or is it just a style cleanup?

AlasdairSwan: Style, I think it makes sense for the labels to be beside the corresponding inputs.

jimabramson:
:+1:

programs/static/js/views/program_details_view.js
```diff
@@ -39,14 +39,15 @@ define([
             },
 
             postRender: function() {
-                var courses = this.model.get('course_codes');
+                var courses = this.model.get( 'course_codes' );
 
                 _.each( courses, function( course ) {
-                    var title = course.key + 'Course';
+                    var title = course.key + 'Course',
```
rlucioni: What's going on with this title? Should there be a space between the course key and `Course`?

AlasdairSwan: This is actually a key for the views so ```this.courseKeyCourse``` will reference said course's view.

rlucioni:
:+1:

PR #50, opened 2015-11-20
Alasdair/ecom 2913 revised app architecture

AlasdairSwan:
@rlucioni @jimabramson please review this PR.

### DIFF
--- a/docs/programs.rst
+++ b/docs/programs.rst
@@ -2,3 +2,20 @@ Programs
 ========
 
 Programs are collections of courses codes offered together. Each "course code" in a Program consists of a collection of course runs, specific instances of a course which a student may take on their way to completing a Program. Together with rules governing which courses may be added to a Program, Programs can be used to support product lines like edX's XSeries offering.
+
+Authoring App
+-----------------
+This application is for creating and managing groupings of courses, which we refer to as **programs**.
+
+
+You may notice that there are 2 versions of the application (core & dev). The core version includes the ability to create new programs and view existing ones. The dev version extends this to also include a programs list view. So the contents of each app are:
+
++----------+---------------------+----------------------+-------------------+
+|          | Create Program View | Program Details View | Program List View |
++==========+=====================+======================+===================+
+| Core App |  |checkmark|        |  |checkmark|         |                   |
++----------+---------------------+----------------------+-------------------+
+| Dev App  |  |checkmark|        |  |checkmark|         |  |checkmark|      |
++----------+---------------------+----------------------+-------------------+
+
+.. |checkmark| unicode:: &#x2713;
--- a/programs/static/js/apps/program_admin_app.js
+++ b/programs/static/js/apps/program_admin_app.js
@@ -1,71 +1,14 @@
 (function() {
     'use strict';
 
+    /**
+     * Allows for rapid toggling between app version locally
+     * i.e. just swap out the file name in the require array
+     */
     require([
-            'backbone',
-            'js/models/pagination_model',
-            'js/routers/program_router'
+            'js/views/program_admin_app_view_core'
         ],
-        function( Backbone, ProgramsModel, ProgramRouter ) {
-            var ProgramAdminApp = Backbone.View.extend({
-                el: '.js-program-admin',
-
-                events: {
-                    'click .js-app-click': 'navigate'
-                },
-
-                initialize: function() {
-                    this.model = new ProgramsModel();
-                    this.model.once( 'sync', this.start, this );
-                    this.model.getList();
-                },
-
-                /**
-                 * Navigate to a new page within the app.
-                 *
-                 * Attempts to open the link in a new tab/window behave as the user expects, however the app
-                 * and data will be reloaded in the new tab/window.
-                 *
-                 * @param {Event} event - Event being handled.
-                 * @returns {boolean} - Indicates if event handling succeeded (always true).
-                 */
-                navigate: function (event) {
-                    var url = $(event.target).attr('href').replace( this.app.root, '' );
-
-                    /**
-                     * Handle the cases where the user wants to open the link in a new tab/window.
-                     * event.which === 2 checks for the middle mouse button (https://api.jquery.com/event.which/)
-                     */
-                    if ( event.ctrlKey || event.shiftKey || event.metaKey || event.which === 2 ) {
-                        return true;
-                    }
-
-                    // We'll take it from here...
-                    event.preventDefault();
-
-                    // Process the navigation in the app/router.
-                    if ( url === Backbone.history.getFragment() && url === '' ) {
-                        /**
-                         * Note: We must call the index directly since Backbone
-                         * does not support routing to the same route.
-                         */
-                        this.app.index();
-                    } else {
-                        this.app.navigate( url, { trigger: true } );
-                    }
-                },
-
-                start: function() {
-                    this.app = new ProgramRouter({
-                        model: this.model
-                    });
-                    this.app.start();
-
-                    // Listen for additions/deletions from results
-                    this.model.on( 'change:results', this.model.fetch, this );
-                }
-            });
-
+        function( ProgramAdminApp ) {
             return new ProgramAdminApp();
         }
     );
--- a/programs/static/js/apps/program_admin_app_core.js
+++ b/programs/static/js/apps/program_admin_app_core.js
@@ -1,0 +1,9 @@
+define([
+        'js/views/program_admin_app_view_core'
+    ],
+    function( ProgramAdminApp ) {
+        'use strict';
+
+        return new ProgramAdminApp();
+    }
+);
--- a/programs/static/js/routers/core_app_program_router.js
+++ b/programs/static/js/routers/core_app_program_router.js
@@ -2,22 +2,34 @@ define([
         'backbone',
         'js/views/program_creator_view',
         'js/views/program_details_view',
-        'js/views/program_list_view'
+        'js/views/program_list_view',
+        'js/models/pagination_model',
+        'js/models/program_model'
     ],
-    function( Backbone, ProgramCreatorView, ProgramDetailsView, ProgramListView ) {
+    function( Backbone, ProgramCreatorView, ProgramDetailsView,
+              ProgramListView, ProgramsListModel, ProgramModel ) {
         'use strict';
 
         return Backbone.Router.extend({
-            root: '/author/',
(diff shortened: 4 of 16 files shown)


programs/static/js/routers/dev_app_program_router.js
```diff
@@ -0,0 +1,33 @@
...
+        'js/routers/core_app_program_router',
+        'js/views/program_list_view'
+    ],
+    function( ProgramsListModel, AdminAppRouter, ProgramListView ) {
+        'use strict';
+
+        return AdminAppRouter.extend({
+            routes: {
+                '': 'programList',
+                'home': 'programList',
```
rlucioni: Is defining a route for "home" necessary?

AlasdairSwan: Maybe it isn't actually. For local dev I can just set the data attribute to be /author

programs/static/js/routers/core_app_program_router.js
```diff
@@ -0,0 +1,68 @@
...
+        'js/views/program_list_view',
+        'js/models/pagination_model',
+        'js/models/program_model'
+    ],
+    function( Backbone, ProgramCreatorView, ProgramDetailsView,
+              ProgramListView, ProgramsListModel, ProgramModel ) {
+        'use strict';
+
+        return Backbone.Router.extend({
+            root: '/author/',
```
rlucioni: The Studio index page is served at "/home," courses are edited at "/course," and libraries are edited at "/library." Given that, it seems like the root for this app should be "/program." The dev app can override this if necessary, or we can rejigger the Django routes.

AlasdairSwan: I'm fine with the dev app having the same url structure so will rejigger the Django routes.

programs/static/js/views/program_admin_app_view_core.js
```diff
@@ -0,0 +1,62 @@
...
+        function( Backbone, ProgramRouter ) {
+            return Backbone.View.extend({
+                el: '.js-program-admin',
+
+                events: {
+                    'click .js-app-click': 'navigate'
+                },
+
+                initialize: function() {
+                    var home = this.$el.data('home-url');
```
rlucioni: Would it be appropriate to include logic for processing a second data attribute containing a URL for the Programs API in this PR? Right now, it's assumed to be `/api/v1/programs/`, but we'll want to be able to specify a fully-qualified URL (e.g., `https://programs.edx.org/api/v1/programs/`) via a data attribute.

AlasdairSwan: I'm thinking it may be better to keep that as a separate PR / Jira story for the sake of cleaness.

programs/static/js/routers/core_app_program_router.js
rlucioni: Cool, let me know if you need a hand. Changing the root to `/program` also means the routes below can be shortened to `/new` and `/:id`.

AlasdairSwan: Yep, thanks.

programs/static/js/views/program_admin_app_view_core.js
rlucioni: Okay, I just don't want to forget about it. It seems like part of this rearchitecture, so maybe it'd be appropriate as a subtask of ECOM-2913.

AlasdairSwan: That works.

programs/static/js/apps/program_admin_app.js
```diff
@@ -1,71 +1,14 @@
...
 
+    /**
+     * Allows for rapid toggling between app version locally
+     * i.e. just swap out the file name in the require array
+     */
     require([
-            'backbone',
-            'js/models/pagination_model',
-            'js/routers/program_router'
+            'js/views/program_admin_app_view_dev'
```
jimabramson: is this the right default value for this?

programs/static/js/apps/program_admin_app_core.js
```diff
@@ -0,0 +1,9 @@
+define([
+        'js/views/program_admin_app_view_core'
+    ],
+    function( ProgramAdminApp ) {
+        'use strict';
+
+        return new ProgramAdminApp();
+    }
+);
```
jimabramson: any reason app_core and app_dev have different structures?

programs/static/js/routers/dev_app_program_router.js
```diff
@@ -0,0 +1,32 @@
...
+    ],
+    function( ProgramsListModel, AdminAppRouter, ProgramListView ) {
+        'use strict';
+
+        return AdminAppRouter.extend({
+            routes: {
+                '': 'programList',
+                'new': 'programCreator',
+                ':id': 'programDetails'
+            },
```
jimabramson: very minor, but is it possible / sane to append the programList to the parent's routes instead of duplicating?

jimabramson:
@AlasdairSwan code looks good AFAICT, but the commit message needs to be fixed, and a request would be for some brief documentation (for future maintainers) that explains why things have been factored this way.

programs/urls.py
```diff
@@ -60,7 +60,7 @@
 
     urlpatterns += [
         # Drops into the Programs authoring app, which handles its own routing.
-        url(r'^author/*', author_view, name='author'),
+        url(r'^program/*', author_view, name='author'),
```
jimabramson: maybe we should rename this from author to admin_dev or something similar, to be consistent with the underlying renames?

programs/static/js/apps/program_admin_app_core.js
AlasdairSwan: Because mistakes were made (by me) :)

AlasdairSwan: Mistakes which have since been rectified.

AlasdairSwan:
@jimabramson re. your comment around brief documentation. Where would said documentation live? I am assuming it is explaining the core vs dev app approach.

rlucioni:
@AlasdairSwan I'd put that documentation in programs/docs, so that it ends up [here](https://open-edx-programs-guide.readthedocs.org/en/latest/).

jimabramson:
@AlasdairSwan yes that is the doc topic I meant.  I would have said to put inline doc, but agree with @rlucioni's better idea to use programs/docs, since the thing you'll document is really spread across many files.

programs/static/js/routers/dev_app_program_router.js
AlasdairSwan: It would be perfectly sane if Backbone would allow such an obvious thing easily (which I have just discovered it does not). Hopefully I can come back to this later and clean up if I find a solution, but I don't want to spin my wheels too long for such a minor update.

AlasdairSwan:
@jimabramson @rlucioni I have made the relevant changes, can you please give it one more look over and let me know if it's good to go?

docs/programs.rst
```diff
@@ -2,3 +2,9 @@ Programs
 ========
 
 Programs are collections of courses codes offered together. Each "course code" in a Program consists of a collection of course runs, specific instances of a course which a student may take on their way to completing a Program. Together with rules governing which courses may be added to a Program, Programs can be used to support product lines like edX's XSeries offering.
+
+Local Development
+-----------------
+
+You may notice that there are 2 versions of the app (core & dev).
+The core version includes the ability to create new programs and view existing onces. The dev version extends this to also include a programs list view.
```
rlucioni: s/onces/ones.

docs/programs.rst
```diff
@@ -2,3 +2,9 @@ Programs
 ========
 
 Programs are collections of courses codes offered together. Each "course code" in a Program consists of a collection of course runs, specific instances of a course which a student may take on their way to completing a Program. Together with rules governing which courses may be added to a Program, Programs can be used to support product lines like edX's XSeries offering.
+
+Local Development
+-----------------
+
+You may notice that there are 2 versions of the app (core & dev).
```
rlucioni: As someone from the open source coming to this with no prior experience, I would appreciate knowing what "the app" is. I'd rename this subsection "Authoring App" and add a brief explanation of what the app is for before explaining the differences between the two versions.

programs/static/js/views/program_admin_app_view_core.js
rlucioni: I've created https://openedx.atlassian.net/browse/ECOM-3034 to track this.

rlucioni:
Looking good @AlasdairSwan, I just made a few minor comments about documentation. Rebasing onto master once https://github.com/edx/programs/pull/54 lands should get your build passing again.

AlasdairSwan:
@rlucioni now with better documentation, can you please take another look?

docs/programs.rst
```diff
@@ -2,3 +2,20 @@ Programs
 ========
 
 Programs are collections of courses codes offered together. Each "course code" in a Program consists of a collection of course runs, specific instances of a course which a student may take on their way to completing a Program. Together with rules governing which courses may be added to a Program, Programs can be used to support product lines like edX's XSeries offering.
+
+Authoring App
+-----------------
+This application is for creating and managing groupings of courses, which we refer to as **programs**.
+
+
+You may notice that there are 2 versions of the application (core & dev). The core version includes the ability to create new programs and view existing onces. The dev version extends this to also include a programs list view. So the contents of each app are;
```
rlucioni: Minor, but it behooves me to point out that this semicolon should really be a colon.

rlucioni:
:+1:

docs/programs.rst
```diff
@@ -2,3 +2,20 @@ Programs
 ========
 
 Programs are collections of courses codes offered together. Each "course code" in a Program consists of a collection of course runs, specific instances of a course which a student may take on their way to completing a Program. Together with rules governing which courses may be added to a Program, Programs can be used to support product lines like edX's XSeries offering.
+
+Authoring App
+-----------------
+This application is for creating and managing groupings of courses, which we refer to as **programs**.
+
+
+You may notice that there are 2 versions of the application (core & dev). The core version includes the ability to create new programs and view existing onces. The dev version extends this to also include a programs list view. So the contents of each app are:
```
jimabramson: s/onces/ones

jimabramson:
:+1:  but is that the intended commit message?

AlasdairSwan:
Yes, I like to list the main parts of the PR to help me remember what I did.

PR #51, opened 2015-12-01
Set up Programs API for CORS

rlucioni:
The Programs API must allow CORS in order for the authoring app to function in Studio. ECOM-2598.

@jimabramson please review. @AlasdairSwan FYI.

### DIFF
--- a/programs/static/js/models/program_model.js
+++ b/programs/static/js/models/program_model.js
@@ -53,7 +53,7 @@ define([
 
             getConfig: function( options ) {
                 var patch = options && options.patch ? true : false,
-                    params = patch ? this.get('id') : '',
+                    params = patch ? this.get('id') + '/' : '',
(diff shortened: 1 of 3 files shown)


programs/static/js/models/program_model.js
```diff
@@ -49,7 +49,7 @@ define([
 
             getConfig: function( options ) {
                 var patch = options && options.patch ? true : false,
-                    params = patch ? this.get('id') : '',
+                    params = patch ? this.get('id') + '/' : '',
```
rlucioni: The [W3C CORS spec](http://www.w3.org/TR/cors/#resource-requests) doesn't allow user agents to follow redirects on CORS requests. Appending this slash prevents Django from responding with a 302 when requests to the Programs API are made.

jimabramson:
:+1:

PR #52, opened 2015-12-01
Use r.js' wrapShim option

rlucioni:
Fixes post-build issues when using an AMD-aware version of Backbone, but using shim config for scripts that depend on Backbone (e.g., Backbone.Validation). ECOM-2598.


AlasdairSwan:
Nice :+1:

PR #53, opened 2015-12-01
Clean up Makefile and add new targets

rlucioni:
New targets are helpful when developing the authoring app for use with external applications such as Studio.

@jimabramson please review. FYI @AlasdairSwan.


jimabramson:
:+1:

PR #56, opened 2015-12-02
Authenticate JS client id_token endpoint.

jimabramson:
ECOM-2440

With this change, the backbone models using the Programs API will attempt to seamlessly recover from 401 errors by initiating the OIDC authorization code flow, then retrying the original API request (once).

@rlucioni @AlasdairSwan please review.

NOTE: the custom pagination_model module is not supported.  This is only used in the 'dev' version of the author app at present.  I think it would make sense implement support for this as part of ECOM-2774.


programs/static/js/test/specs/auto_auth_model_spec.js
```diff
@@ -0,0 +1,130 @@
...
+                // stand-in for the error callback a caller might pass with options to Backbone.Model.sync
+                testErrorCallback = jasmine.createSpy();
+
+                fakeAjaxDeferred = $.Deferred();
+                spyOn( $, 'ajax' ).and.returnValue( fakeAjaxDeferred );
+                return fakeAjaxDeferred;
+
+            });
+
+            function spyOnBackboneSync( status ) {
```
AlasdairSwan: This function and the 2 that follow do not have var declarations.
Nitpick around syntax. For consistency I would prefer to maintain the format of 
``` var spyOnBackboneSync = function( status ) {...```
(but with the vars actually defined at the top of the describe function on line 12)

Comment (jimabramson) [inline programs/static/js/test/specs/auto_auth_model_spec.js]: @AlasdairSwan fixed - didn't change/move the local variable declarations, not sure if you were asking for that.

File: programs/static/js/test/specs/auto_auth_model_spec.js [hunk 2]
```diff
@@ -0,0 +1,130 @@
...
+                // make sure Backbone.sync was called with custom error handling
+                expect( Backbone.sync.calls.count() ).toEqual(1);
+                params = _.object( ['method', 'model', 'options'], Backbone.sync.calls.mostRecent().args );
+                expect( params.method ).toEqual( 'GET' );
+                expect( params.model ).toEqual( dummyModel );
+                expect( params.options.error ).not.toEqual( testErrorCallback );
+                return params;
+            };
+
+            var checkAuthAttempted = function(isExpected) {
```
AlasdairSwan: Cool, but I was also wanting the var declarations at the top of the describe function because this is a JS best practice. A nitpick I suppose but a good habit to get into to avoid hoisting.

AlasdairSwan:
Looks good :+1:

rlucioni:
@jimabramson looks good to me, :+1: . I'd like to verify that this works from Studio before merging.

rlucioni:
https://github.com/edx/programs/pull/57 should help with that.

jimabramson:
As we've uncovered a lot of issues with oauth / cors / csrf under this approach, it will be revised to use a new custom endpoint to fetch tokens directly from studio.  will update when ready for re-review.

jimabramson:
This branch changed somewhat - to keep things clean, going to close this and open a new PR.

PR #57, opened 2015-12-03
Allow restructured authoring app to be bundled

rlucioni:
Also initialize program model with a CORS-friendly URL (no redirection required).

@jimabramson this will help fix the issues we were seeing earlier. @AlasdairSwan FYI.


jimabramson:
:+1:    this definitely did the trick locally.

PR #59, opened 2015-12-07
Automatically authenticate client using id token endpoint.

jimabramson:
ECOM-2440

Changed direction here, to avoid nasty issues with CORS and CSRF.  Now, we use a dedicated endpoint to request an id token for API requests whenever we need one, and save its value in a new model in the client app.

Depends on a new endpoint in Studio, which merged in https://github.com/edx/edx-platform/pull/10862.  Note that this works fine when the app is embedded studio; using the development version currently requires modifying Studio to whitelist the id_token endpoint with CORS. 

@rlucioni @AlasdairSwan please review.

### DIFF
--- a/programs/static/js/models/api_config_model.js
+++ b/programs/static/js/models/api_config_model.js
@@ -1,0 +1,15 @@
+define([
+        'backbone'
+    ],
+    function( Backbone ) {
+        'use strict';
+
+        return Backbone.Model.extend({
+            defaults: {
+                'base_url': 'http://127.0.0.1:8009/api/v1/',
+                'auth_url' : 'http://127.0.0.1:8001/programs/id_token/',
+                'id_token' : ''
+            }
+        });
+    }
+);
--- a/programs/static/js/models/auto_auth_model.js
+++ b/programs/static/js/models/auto_auth_model.js
@@ -1,0 +1,84 @@
+define([
+        'backbone',
+        'jquery',
+        'js/utils/api_config'
+    ],
+    function( Backbone, $, apiConfig ) {
+        'use strict';
+
+        return Backbone.Model.extend({
+
+            /**
+             * Override Backbone.sync to seamlessly attempt (re-)authentication when necessary.
+             *
+             * If a 401 error response is encountered while making a request to the Programs,
+             * API, this wrapper will attempt to request an id token from a custom endpoint
+             * via AJAX.  Then the original request will be retried once more.
+             *
+             * Any other response than 401 on the original API request, or any error occurring
+             * on the retried API request (including 401), will be handled by the base sync
+             * implementation.
+             *
+             */
+            sync: function( method, model, options ) {
+
+                var oldError = options.error;
+
+                this._setHeaders( options );
+
+                options.error = function(xhr, textStatus, errorThrown) {
+                    if (xhr && xhr.status === 401) {
+                        // attempt auth and retry
+                        this._updateToken(function() {
+                            // restore the original error handler
+                            options.error = oldError;
+                            delete options.xhr;  // remove the failed (401) xhr from the last try.
+
+                            // update authorization header
+                            this._setHeaders( options );
+
+                            Backbone.sync.call(this, method, model, options);
+                        }.bind(this));
+                    } else if (oldError) {
+                        // fall back to the original error handler
+                        oldError.call(this, xhr, textStatus, errorThrown);
+                    }
+                }.bind(this);
+                return Backbone.sync.call(this, method, model, options);
+            },
+
+            /**
+             * Fix up headers on an imminent AJAX sync, ensuring that the JWT token is enclosed
+             * and that credentials are included when the request is being made cross-domain.
+             */
+            _setHeaders: function( ajaxOptions ) {
+                ajaxOptions.headers = _.extend ( ajaxOptions.headers || {}, {
+                    Authorization: 'JWT ' + apiConfig.get( 'id_token' )
+                });
+                ajaxOptions.xhrFields = _.extend( ajaxOptions.xhrFields || {}, {
+                    withCredentials: true
+                });
+            },
+
+            /**
+             * Fetch a new id token from the configured endpoint, update the api config,
+             * and invoke the specified callback.
+             */
+            _updateToken: function( success ) {
+
+                $.ajax({
+                    url: apiConfig.get('auth_url'),
+                    xhrFields: {
+                        withCredentials: true
+                    },
+                    crossDomain: true
+                }).done(function ( data ) {
+                    // save the newly-retrieved id token
+                    apiConfig.set( 'id_token', data.id_token );
+                }).done( success );
+
+            }
+
+        });
+    }
+);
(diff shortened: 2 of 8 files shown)


programs/static/js/models/api_config_model.js
```diff
@@ -0,0 +1,23 @@
...
+                'auth_url' : 'http://127.0.0.1:8001/programs/id_token/',
+                'id_token' : ''
+            }
+        });
+
+        if (instance === undefined) {
+            instance = new ApiConfigModel();
+        }
+
+        return instance;
```
jimabramson: Note: singleton!  appears to be working great (all models are sharing an id token).   This would seem to be a good place to read/write other globally-useful config values, avoiding the need to pass lots of arguments around between layers.

programs/static/js/models/auto_auth_model.js
```diff
@@ -0,0 +1,87 @@
...
+    ],
+    function( Backbone, $, ApiConfigModel ) {
+        'use strict';
+
+        return Backbone.Model.extend({
+
+            /**
+             * Override Backbone.sync to seamlessly attempt (re-)authentication when necessary.
+             *
+             * If a 401 error response is encountered while making a request to the Programs,
```
jimabramson: Oops.   updating.

programs/static/js/models/program_model.js
```diff
@@ -56,9 +57,7 @@ define([
                     params = patch ? this.get('id') + '/' : '',
                     config = _.extend({ validate: true, parse: true }, {
                         type: patch ? 'PATCH' : 'POST',
-                        url: this.urlRoot + params,
-                        // The API requires a CSRF token for all POST requests using session authentication.
-                        headers: {'X-CSRFToken': $.cookie('programs_csrftoken')},
```
jimabramson: N.B. no longer necessary when using token authentication.

programs/static/js/models/api_config_model.js
```diff
@@ -0,0 +1,23 @@
+define([
+        'backbone'
+    ],
+    function( Backbone ) {
+        'use strict';
+
+        var instance;
+
+        var ApiConfigModel = Backbone.Model.extend({
```
rlucioni: Style nitpick: fold this var declaration into the one above:
```javascript
var instance,
    apiConfigModel = Backbone.Model.extend({
    });
```

programs/static/js/models/api_config_model.js
rlucioni: Interesting. I'd be curious to see what @AlasdairSwan thinks of this. I like the prospect of not having to pass options across several layers of the app. Case in point: [my first pass](https://github.com/edx/programs/compare/renzo/url-from-dom) at grabbing the Programs API URL from the DOM.

programs/static/js/models/pagination_model.js
```diff
@@ -1,13 +1,14 @@
...
         'backbone',
         'jquery',
+        'js/models/api_config_model',
         'js/collections/programs_collection'
     ],
-    function( Backbone, $, ProgramsCollection ) {
+    function( Backbone, $, ApiConfigModel, ProgramsCollection ) {
         'use strict';
 
         return Backbone.Model.extend({
```
rlucioni: This model makes requests to get a list of programs, so it seems like it should also extend the AutoAuthModel.

rlucioni:
Looks nice @jimabramson. I've left comments inline.

Since it looks like this PR will change our approach to managing configuration, I'll put [my WIP](https://github.com/edx/programs/compare/renzo/url-from-dom) on hold until this lands.

programs/static/js/models/auto_auth_model.js
```diff
@@ -0,0 +1,85 @@
...
+             *
+             * Any other response than 401 on the original API request, or any error occurring
+             * on the retried API request (including 401), will be handled by the base sync
+             * implementation.
+             *
+             */
+            sync: function( method, model, options ) {
+
+                var oldError = options.error,
+                    self = this;
```
AlasdairSwan: We shouldn't have to alias this, should be able to use _.bind(this) instead (which I find cleaner and less confusing)

programs/static/js/models/auto_auth_model.js
```diff
@@ -0,0 +1,85 @@
...
+                            // update authorization header
+                            self._setHeaders( options );
+
+                            Backbone.sync.call(self, method, model, options);
+                        });
+                    } else if (oldError) {
+                        // fall back to the original error handler
+                        oldError.call(self, xhr, textStatus, errorThrown);
+                    }
+                };
```
AlasdairSwan: It seems like this should be moved to its own function, which would make it easier to extend the model is we wanted to later.

AlasdairSwan: Actually, it seems like with the need for the oldError that may not be great so it's fine here.

programs/static/js/models/auto_auth_model.js
```diff
@@ -0,0 +1,85 @@
...
+
+            /**
+             * Fix up headers on an imminent AJAX sync, ensuring that the JWT token is enclosed
+             * and that credentials are included when the request is being made cross-domain.
+             */
+            _setHeaders: function( ajaxOptions ) {
+                ajaxOptions.headers = _.extend ( ajaxOptions.headers || {}, {
+                    Authorization: 'JWT ' + ApiConfigModel.get( 'id_token' )
+                });
+                ajaxOptions.xhrFields = _.extend( ajaxOptions.xhrFields || {}, { withCredentials: true } );
```
AlasdairSwan: Nitpick: you took a new line for the second object in the previous extend but not this one.

programs/static/js/models/api_config_model.js
AlasdairSwan: This is interesting, but I am slightly nervous that you cannot actually extend the ApiConfigModel, just an instance of it. I'm not entirely sure if that's an issue or not to be honest.

AlasdairSwan: @jimabramson what do you think about moving this across 2 files? So there is the api_config_model.js and then programs_app_config.js which is just the instance part (and this is what would be called throughout the app). It may be overkill, but it makes it clearer what is going on and feels a little more Backbonesque. Would that work?

jimabramson: that will work but I'm not seeing much use for the first part once completed.  it'd just be a plain Backbone model with no customizations in it, right?

jimabramson:
@AlasdairSwan @rlucioni i've pushed another fixup commit addressing all the feedback.

programs/static/js/models/pagination_model.js
jimabramson: @rlucioni note I've done this for consistency's sake, but it won't work.  The pagination model doesn't use Backbone.sync.  As I mentioned in the previous instance of this PR, I think we should address that problem when we update the pagination implementation (separate story).  FWIW this model is only used in the "development" version of the app.

jimabramson:
last commit was borked, fixed now

AlasdairSwan:
Nice, once tests pass and commits are squashed :+1:

rlucioni:
:+1:

PR #62, opened 2015-12-10
Allow nested writes during Program updates.

jimabramson:
ECOM-2210

This allows program course codes and run modes to be updated as part of a Programs PATCH request.

Note that this does not support creating program course codes based on course codes not yet known to the system.  This PR was getting big, so I plan to address that in a second, separate PR.

@rlucioni  I'm aware of a coverage gap that needs to be closed, but this is otherwise ready for review.

### DIFF
--- a/programs/apps/api/serializers.py
+++ b/programs/apps/api/serializers.py
@@ -6,11 +6,57 @@ should be created here. As the API evolves, serializers may become more
 specific to a particular version of the API. In this case, the serializers
 in question should be moved to versioned sub-package.
 """
+from django.core.exceptions import ValidationError
 from django.utils.translation import ugettext as _
-from rest_framework import serializers
-
+from rest_framework import fields, exceptions, serializers
 
 from programs.apps.programs import models, constants
+
+
+class NestedWriteableSerializer(serializers.ListSerializer):
+    """
+    Reusable implementation of updatable nested lists.
+
+    In order to use this serializer class, you must:
+      a) specify it as the list_serializer_class for the serializer being nested
+      b) define classmethod `unique_attrs(obj)` on the serializer being nested.  That
+         method should return a tuple of values that are guaranteed to be unique
+         for any child relative to the parent, and it should work with either
+         mapping objects (keys) or model instances (attributes).  See the examples
+         defined in this file.
+    """
+
+    def update(self, instance, validated_data):
+
+        def _key(obj):
+            """
+            Use unique_attrs to compare across nested json dicts and model
+            instances.
+            """
+            return self.child.unique_attrs(obj)
+
+        db_objs = {_key(obj): obj for obj in instance}
+        req_objs = {_key(obj): obj for obj in validated_data}
+
+        # Perform creations and updates.
+        ret = []
+
+        for key, req_obj in req_objs.items():
+            db_obj = db_objs.get(key)
+            if db_obj is None:
+                # create the thing
+                ret.append(self.child.create(req_obj))
+            else:
+                # update the thing
+                ret.append(self.child.update(db_obj, req_obj))
+
+        # Perform deletions.
+        for key, db_obj in db_objs.items():
+            if key not in req_objs:
+                # delete the thing
+                db_obj.delete()
+
+        return ret
 
 
 class OrganizationSerializer(serializers.ModelSerializer):
@@ -38,6 +84,21 @@ class ProgramCourseRunModeSerializer(serializers.ModelSerializer):
     class Meta(object):  # pylint: disable=missing-docstring
         model = models.ProgramCourseRunMode
         fields = ('course_key', 'mode_slug', 'sku', 'start_date', 'run_key')
+        list_serializer_class = NestedWriteableSerializer
+
+    @classmethod
+    def unique_attrs(cls, obj):
+        """
+        For use with `NestedWriteableSerializer.update`
+        """
+        try:
+            if isinstance(obj, models.ProgramCourseRunMode):
+                return obj.course_key, obj.mode_slug, obj.sku
+            else:
+                return obj['course_key'], obj['mode_slug'], obj.get('sku')
+        except (AttributeError, KeyError) as exc:
+            # avoid errors when working with incomplete/invalid nested data
+            raise exceptions.ValidationError(exc.message)
 
 
 class CourseCodeSerializer(serializers.ModelSerializer):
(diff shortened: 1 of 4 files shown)


programs/apps/api/v1/tests/test_views.py
```diff
@@ -380,6 +380,181 @@ def test_view_with_nested(self):
...
         )
 
+    def test_update_course_codes(self):
+        """
+        Ensure that nested program course codes can be updated.
+        """
+        org = OrganizationFactory.create(key="test-org-key")
+        program = ProgramFactory.create()
+        ProgramOrganizationFactory.create(program=program, organization=org)
+        for n in range(3):
```
rlucioni: Is it necessary to create three of these course codes when you only need two, one associated and one not?

programs/apps/api/v1/tests/test_views.py
```diff
@@ -380,6 +380,181 @@ def test_view_with_nested(self):
...
+        # check models (ensure things were saved properly)
+        db_course_codes = ProgramCourseCode.objects.filter(program__id=program.id)
+        self.assertEqual(
+            [u"test-course-key-0", u"test-course-key-2"],
+            sorted([pcc.course_code.key for pcc in db_course_codes])
+        )
+
+    def test_update_run_modes(self):
+        """
+        Ensure that nested program course codes can be updated.
```
rlucioni: s/course codes/run modes

I love that it rhymes.

programs/apps/api/v1/tests/test_views.py
jimabramson: ultimately 3 are needed because the test creates one association (the third), deletes another one (the second), and ignores one (the first)

programs/apps/api/v1/tests/test_views.py
```diff
@@ -380,6 +380,181 @@ def test_view_with_nested(self):
...
+                        },
+                    ],
+                },
+            ],
+        }
+        response = self._make_request(program_id=program.id, admin=True, method='patch', data=patch_data)
+        self.assertEqual(response.status_code, 200)
+
+        # check response data
+        patched_run_modes = json.loads(response.content)["course_codes"][0]["run_modes"]
```
rlucioni: Minor: here and throughout, you should be able to access `response.data` to access the deserialized data.

programs/apps/api/serializers.py
```diff
@@ -75,7 +209,7 @@ class Meta(object):  # pylint: disable=missing-docstring
         read_only_fields = ('id', 'created', 'modified')
 
     organizations = ProgramOrganizationSerializer(many=True, source='programorganization_set')
-    course_codes = ProgramCourseCodeSerializer(many=True, read_only=True, source='programcoursecode_set')
+    course_codes = ProgramCourseCodeSerializer(many=True, source='programcoursecode_set', required=False)
```
rlucioni: This is tangentially related to the changes made here, but I've just realized that our use of nested relationships like this one might be responsible for the Program list view's poor performance (CC @AlasdairSwan).

@jimabramson correct me if I'm wrong, but when someone requests a list of `n` Programs, this serializer first fetches all programs from the database, then for each program, fetches its course codes. This requires an additional trip to the database for each program, resulting in `n+1` calls to the database. It looks like the `ProgramCourseCodeSerializer` is also vulnerable to this; it fetches all course codes from the database, then for each course code, we return to the database to fetch run modes.

We should be able to address this by calling [prefetch_related()](https://docs.djangoproject.com/en/1.9/ref/models/querysets/#prefetch-related) in our viewsets' `get_queryset()`, as detailed [here](https://ses4j.github.io/2015/11/23/optimizing-slow-django-rest-framework-performance/).

programs/apps/api/v1/tests/test_views.py
```diff
@@ -380,6 +380,181 @@ def test_view_with_nested(self):
...
+        for course_code, response_course_code in zip(course_codes, response_course_codes):
+            self.assertEqual(response_course_code["key"], course_code.key)
+            self.assertEqual(response_course_code["organization"]["key"], course_code.organization.key)
+            self.assertEqual(len(response_course_code["run_modes"]), 2)
+
+            # check db values. the following call should throw DoesNotExist if things didn't go as expected
+            db_program_course_code = ProgramCourseCode.objects.get(program=program, course_code=course_code)
+            run_modes = sorted(db_program_course_code.run_modes.all(), key=lambda o: o.course_key)
+            for i, run_mode in enumerate(run_modes):
+                self.assertEqual("course-v1:org+{0}+run-{1}".format(course_code.key, i), run_mode.course_key)
```
rlucioni: Minor: you don't need to specify indices when performing string formatting.

programs/apps/api/serializers.py
```diff
@@ -38,6 +85,17 @@ class ProgramCourseRunModeSerializer(serializers.ModelSerializer):
...
+
+    @classmethod
+    def unique_attrs(cls, obj):
+        """
+        For use with `NestedWriteableSerializer.update`
+        """
+        if isinstance(obj, models.ProgramCourseRunMode):
+            return obj.course_key, obj.mode_slug, obj.sku
+        else:
+            return obj["course_key"], obj["mode_slug"], obj.get('sku')
```
rlucioni: Also minor, but I feel compelled to say something: sticking to one quote style (e.g., single quotes) would be nice!

programs/apps/api/v1/tests/test_views.py
rlucioni: @jimabramson understood. In that case, maybe it'd be a good idea to add an assertion validating that the second course code (`test-course-key-1`) is deleted as expected.

rlucioni: Never mind, you do that implicitly by checking all course codes associated with the program.

programs/apps/api/serializers.py
jimabramson: good call @rlucioni.  However I'd like to make that change in a followup PR, and try to keep the scope of this one from creeping too much.

jimabramson: created https://openedx.atlassian.net/browse/ECOM-3173

rlucioni: @jimabramson cool, thank you. Maybe we can discuss at our next grooming.

jimabramson:
@rlucioni the most recent fixup commit addresses your comments; i'm getting one more ready to address the test coverage.

rlucioni:
:+1: once coverage restored.

jimabramson:
@rlucioni I've reorganized some of the validation logic and added tests for various invalid inputs in a second fixup commit.  Please give another look when you have a moment.

programs/apps/api/v1/tests/test_views.py
```diff
@@ -380,6 +376,255 @@ def test_view_with_nested(self):
...
+    def test_invalid_nested_run_modes(self, invalid_mode):
+        """
+        Ensure that invalid nested run mode data causes a 400 response.
+        """
+        org = OrganizationFactory.create()
+        program = ProgramFactory.create()
+        ProgramOrganizationFactory.create(program=program, organization=org)
+        course_code = CourseCodeFactory.create(organization=org)
+
+        # associate two course modes with two run modes each, in one PATCH request.
```
rlucioni: This comment seems incorrect. The PATCH request is used to attempt association of a single course code with an invalid run mode.

programs/apps/api/v1/tests/test_views.py
```diff
@@ -380,6 +376,255 @@ def test_view_with_nested(self):
...
+        Ensure that nested program course codes and run modes can be correctly
+        created during updates.
+        """
+        org = OrganizationFactory.create()
+        program = ProgramFactory.create()
+        ProgramOrganizationFactory.create(program=program, organization=org)
+
+        course_codes = [CourseCodeFactory.create(organization=org, key='test-cc-{}'.format(i)) for i in range(2)]
+
+        # associate two course modes with two run modes each, in one PATCH request.
```
rlucioni: s/course modes/course codes

rlucioni:
:+1: once comments revised.

PR #63, opened 2015-12-10
ECOM 2764 course data input fields

AlasdairSwan:
@rlucioni @jimabramson please review. Not that once the APIs and all of the other  sub-tasks have been completed the tests will be more fully updated so increase coverage.

### DIFF
--- a/programs/static/js/models/course_model.js
+++ b/programs/static/js/models/course_model.js
@@ -1,0 +1,25 @@
+define([
+        'backbone',
+        'jquery',
+        'js/utils/api_config',
+        'js/models/auto_auth_model',
+        'jquery-cookie'
+    ],
+    function( Backbone, $, apiConfig, AutoAuthModel ) {
+        'use strict';
+
+        return AutoAuthModel.extend({
+
+            validation: {
+                key: {
+                    required: true,
+                    maxLength: 64
+                },
+                display_name: {
+                    required: true,
+                    maxLength: 128
+                }
+            }
+        });
+    }
+);
--- a/programs/static/js/views/course_details_view.js
+++ b/programs/static/js/views/course_details_view.js
@@ -1,15 +1,18 @@
 define([
         'backbone',
+        'backbone.validation',
         'jquery',
         'underscore',
+        'js/models/course_model',
         'js/models/pagination_model',
         'js/models/program_model',
         'js/views/course_run_view',
         'text!templates/course_details.underscore',
-        'gettext'
+        'gettext',
+        'js/utils/validation_config'
     ],
-    function( Backbone, $, _, ProgramsModel, ProgramModel,
-              CourseRunView, ListTpl ) {
+    function( Backbone, BackboneValidation, $, _, CourseModel, ProgramsModel,
+              ProgramModel, CourseRunView, ListTpl ) {
         'use strict';
 
         return Backbone.View.extend({
@@ -18,8 +21,8 @@ define([
             className: 'course-details',
 
             events: {
-                'change .js-course-select': 'setCourse',
-                'click .js-remove-course': 'destroy'
+                'click .js-remove-course': 'destroy',
+                'click .js-select-course': 'setCourse'
             },
 
             tpl: _.template( ListTpl ),
@@ -28,30 +31,36 @@ define([
             courseRuns: [],
 
             initialize: function( options ) {
-                this.model = new ProgramModel();
+
+                this.programModel = new ProgramModel();
+                this.model = new CourseModel();
+                Backbone.Validation.bind( this );
                 this.$parentEl = $( this.parentEl );
                 this.programStatus = options.programStatus;
 
                 // Data passed in because it is not an actual model
                 if ( options.data ) {
-                    this.model.set( options.data );
+                    this.programModel.set( options.data );
                 } else {
-                    this.courseList = options.courseList.toJSON();
                     this.setDefaultModel();
                 }
 
-                this.model.on('change:display_name', this.update, this);
+                //this.programModel.on('change:display_name', this.update, this);
                 this.render();
             },
 
             render: function() {
-                this.$el.html( this.tpl( this.model.toJSON() ) );
+                var data = _.extend( this.programModel.toJSON(), {
+                    cid: this.programModel.cid
+                });
+
+                this.$el.html( this.tpl( data ) );
                 this.$parentEl.append( this.$el );
                 this.postRender();
             },
 
             postRender: function() {
-                var runs = this.model.get('run_modes');
+                var runs = this.programModel.get('run_modes');
 
                 if ( runs && runs.length > 0 ) {
                     this.addCourseRuns();
@@ -60,12 +69,12 @@ define([
 
             addCourseRuns: function() {
                 // Create run views
-                var runs = this.model.get('run_modes'),
+                var runs = this.programModel.get('run_modes'),
                     $runsContainer = this.$el.find('.js-course-runs');
 
                 _.each( runs, function( run ) {
                     var runView = new CourseRunView({
-                        model: this.model,
+                        model: this.programModel,
                         data: run,
                         $parentEl: $runsContainer
                     });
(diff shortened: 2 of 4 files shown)


programs/static/js/models/course_model.js
```diff
@@ -0,0 +1,79 @@
...
+                        success: _.bind( this.saveSuccess, this ),
+                        error: _.bind( this.saveError, this )
+                    });
+
+                if ( patch ) {
+                    config.data = JSON.stringify( options.update ) || this.attributes;
+                }
+
+                return config;
+            },
```
jimabramson: this looks familiar.  should it be moved into a base class?

AlasdairSwan: Possibly, this was a copy & paste job and the other subtasks are going to tie it up to an API so it's kind of a WIP PR. I suggest waiting until the full implementation is finished. I could just delete the unused parts of this model for now if that would be cleaner?

jimabramson: not a major issue for me if this is still WIP, we can clean it up when dependencies land & are integrated.   by the same token if it's not relevant to the changes you want reviewed, please feel free to remove it :-)

AlasdairSwan:
@jimabramson I cleaned up the model as discussed.

programs/static/templates/course_details.underscore
```diff
@@ -22,11 +22,24 @@
...
-    <select class="js-course-select">
-        <option><%= gettext('Select a course') %></option>
-        <% _.each( courseList, function( course ) { %>
-            <option value="<%- course.id %>"><%- course.name %></option>
-        <% }); %>
-    </select>
+    <form class="form js-course-form">
+        <fieldset class="form-group">
+            <div class="field">
+                <label class="field-label" for="course-key-<%- cid %>"><%= gettext('Course Key') %></label>
```
rlucioni: Relatively minor, but this should be a "Course Code" instead of a "Course Key." I'd change the class name to reflect that, too.

programs/static/templates/course_details.underscore
```diff
@@ -22,11 +22,24 @@
...
+                <input id="course-key-<%- cid %>" class="field-input input-text course-key" name="course_key" maxlength="255" required>
+                <div class="field-message">
+                    <span class="field-message-content"></span>
+                </div>
+            </div>
+            <div class="field">
+                <label class="field-label" for="course-key-<%- cid %>"><%= gettext('Course Title') %></label>
```
rlucioni: The `for` parameter should be `course-title`, right?

programs/static/templates/course_details.underscore
```diff
@@ -22,11 +22,24 @@
...
+                <input id="course-title-<%- cid %>" class="field-input input-text course-title" name="course_title" maxlength="255" required>
+                <div class="field-message">
+                    <span class="field-message-content"></span>
+                </div>
+            </div>
+            <button class="btn btn-primary js-select-course"><%- gettext('Select Course') %></button>
```
rlucioni: s/Select/Add?

programs/static/js/views/course_details_view.js
```diff
@@ -28,30 +31,36 @@ define([
...
             initialize: function( options ) {
-                this.model = new ProgramModel();
+
+                this.programModel = new ProgramModel();
+                this.model = new CourseModel();
+                Backbone.Validation.bind( this );
                 this.$parentEl = $( this.parentEl );
                 this.programStatus = options.programStatus;
 
                 // Data passed in because it is not an actual model
```
rlucioni: Is this still true?

programs/static/templates/course_details.underscore
AlasdairSwan: There is already a button on the page saying 'Add a course'. Probably "Save Course' is better. @marcotuts thoughts?

jimabramson: my $.02 is on 'save'

rlucioni: Save works for me.

jimabramson:
@AlasdairSwan @marcotuts while we're allowing this input as free text i think some help text would be good to explain what these values mean/do.

for the key part, it might be best to stay consistent with Studio's messaging for the same element: "The unique number that identifies your course within your organization, e.g. CS101"  This *should* match the value in course runs in the LMS (though in practice our course run ids are not that clean, so it's not going to be enforced).

for the display name part the import thing is that this value is going to override whatever title might be set for the individual runs of the course, and will be displayed on the xseries progress page and in marketing presentations.  If the course runs' individual titles are changed it won't automatically be reflected here.

programs/static/js/views/course_details_view.js
AlasdairSwan: Yes it is

AlasdairSwan:
@jimabramson @rlucioni I have updated my PR in light of the review comments and added text for the field instructions. @marcotuts please review (and then offer improved versions of) the form field instructional text.

programs/static/js/models/course_model.js
```diff
@@ -0,0 +1,25 @@
...
+    ],
+    function( Backbone, $, apiConfig, AutoAuthModel ) {
+        'use strict';
+
+        return AutoAuthModel.extend({
+
+            validation: {
+                course_code: {
+                    required: true,
+                    maxLength: 255
```
rlucioni: The database restricts this to 64 characters.

programs/static/js/models/course_model.js
```diff
@@ -0,0 +1,25 @@
...
+                },
+                course_title: {
+                    required: true,
+                    maxLength: 255
```
rlucioni: The database restricts this to 128 characters.

programs/static/js/models/course_model.js
```diff
@@ -0,0 +1,25 @@
...
+            validation: {
+                course_code: {
+                    required: true,
+                    maxLength: 64
+                },
+                course_title: {
+                    required: true,
+                    maxLength: 128
+                }
+            }
```
jimabramson: On the server this model is called CourseCode and the fields are 'key' and 'display_name'.  Would recommend staying closer to the same nomenclature if possible.

To work with the recent API changes this model is going to need another field 'organization' which I'm pretty sure needs to be a nested model.  The nested model won't/shouldn't change so you don't have to worry about updates to it, but it needs to be sent along with the course code object for creation/update/deletion.

programs/static/js/views/course_details_view.js
```diff
@@ -28,30 +31,36 @@ define([
...
                 if ( options.data ) {
-                    this.model.set( options.data );
+                    this.programModel.set( options.data );
                 } else {
-                    this.courseList = options.courseList.toJSON();
                     this.setDefaultModel();
                 }
 
-                this.model.on('change:display_name', this.update, this);
+                //this.programModel.on('change:display_name', this.update, this);
```
rlucioni: Should this be commented out?

programs/static/js/models/course_model.js
jimabramson: When creating the organization is always going to match the organization associated with the program so if necessary you can copy its values from there.

programs/static/js/views/course_details_view.js
AlasdairSwan: It's to get test coverage to not be jacked - my next PR after this implements it properly.

rlucioni:
:+1:

programs/static/js/views/course_details_view.js
```diff
@@ -28,30 +31,36 @@ define([
...
+                });
+
+                this.$el.html( this.tpl( data ) );
                 this.$parentEl.append( this.$el );
                 this.postRender();
             },
 
             postRender: function() {
-                var runs = this.model.get('run_modes');
+                var runs = this.programModel.get('run_modes');
```
jimabramson: i'm not sure if i should be ignoring this (for now) but in this and other places, it's implies that run_modes hang directly off the program in the data model, which isn't the case.  a program has many course codes, and each course code has many run modes.

jimabramson:
:+1: after rebasing.   Per discussion with @AlasdairSwan offline, we will deal with data model impedance mismatches in the next task/PR.

AlasdairSwan:
To make progress I am going to just merge this copy because the app is not on production yet and we can make any changes after Sylvia's review in a later PR.

PR #64, opened 2015-12-11
During program updates, create/update course codes on the fly.

jimabramson:
@AlasdairSwan @rlucioni FYI.

This is based on top of #62.  I'm going to return to that one now, but want to get this one on your radar, particularly because it may be useful to unblocking JS work that depends on it.


jimabramson:
this has been rebased on top of the latest updates to #62

rlucioni:
Looks good @jimabramson. I just made a few minor comments. :+1:

PR #65, opened 2015-12-14
Alasdair/ecom 2762 get course runs from lms

AlasdairSwan:
@rlucioni @jimabramson I still need to add tests but in the interests of time if you could review the code so far that would be great.

### DIFF
--- a/programs/static/js/collections/course_runs_collection.js
+++ b/programs/static/js/collections/course_runs_collection.js
@@ -1,0 +1,44 @@
+define([
+        'backbone',
+        'jquery',
+        'js/utils/api_config',
+        'js/collections/auto_auth_collection',
+        'jquery-cookie'
+    ],
+    function( Backbone, $, apiConfig, AutoAuthCollection ) {
+        'use strict';
+
+        return AutoAuthCollection.extend({
+            allRuns: [],
+
+            initialize: function(models, options) {
+                // Ignore pagination and give me everything
+                var orgStr = options.organization.display_name,
+                    queries = '?' + orgStr + '&page_size=1000';
+
+                this.url = apiConfig.get('lmsBaseUrl') + 'api/courses/v1/courses/' + queries;
+            },
+
+            parse: function(data) {
+                this.allRuns = data.results;
+
+                // Because pagination is ignored just set results
+                return data.results;
+            },
+
+            // Adds a run back into the model for selection
+            addRun: function(id) {
+                var courseRun = _.findWhere( this.allRuns, { course_id: id });
+
+                this.create(courseRun);
+            },
+
+            // Removes a run from the model for selection
+            removeRun: function(id) {
+                var courseRun = this.where({course_id: id});
+
+                this.remove(courseRun);
+            }
+        });
+    }
+);
--- a/programs/static/js/models/api_config_model.js
+++ b/programs/static/js/models/api_config_model.js
@@ -6,7 +6,8 @@ define([
 
         return Backbone.Model.extend({
             defaults: {
-                baseUrl: 'http://127.0.0.1:8004/api/v1/',
+                lmsBaseUrl: 'http://127.0.0.1:8001/',
+                programsBaseUrl: 'http://127.0.0.1:8004/',
                 authUrl: 'http://127.0.0.1:8001/programs/id_token/',
                 idToken: ''
             }
--- a/programs/static/js/models/course_model.js
+++ b/programs/static/js/models/course_model.js
@@ -19,6 +19,13 @@ define([
                     required: true,
                     maxLength: 128
                 }
+            },
+
+            defaults: {
+                display_name: false,
+                key: false,
+                organization: [],
+                run_modes: []
             }
         });
     }
--- a/programs/static/js/models/course_run_model.js
+++ b/programs/static/js/models/course_run_model.js
@@ -1,0 +1,9 @@
+define([
+        'backbone'
+    ],
+    function( Backbone ) {
+        'use strict';
+
+        return Backbone.Model.extend({});
+    }
+);
--- a/programs/static/js/utils/auth_utils.js
+++ b/programs/static/js/utils/auth_utils.js
@@ -1,0 +1,86 @@
+define([
+        'jquery',
+        'underscore',
+        'js/utils/api_config'
+    ],
+    function( $, _, apiConfig ) {
+        'use strict';
+
+        var auth = {
+            autoSync: {
+                /**
+                 * Override Backbone.sync to seamlessly attempt (re-)authentication when necessary.
+                 *
+                 * If a 401 error response is encountered while making a request to the Programs,
+                 * API, this wrapper will attempt to request an id token from a custom endpoint
+                 * via AJAX.  Then the original request will be retried once more.
+                 *
+                 * Any other response than 401 on the original API request, or any error occurring
+                 * on the retried API request (including 401), will be handled by the base sync
+                 * implementation.
+                 *
+                 */
+                sync: function( method, model, options ) {
+
+                    var oldError = options.error;
+
+                    this._setHeaders( options );
+
+                    options.error = function(xhr, textStatus, errorThrown) {
+                        if (xhr && xhr.status === 401) {
+                            // attempt auth and retry
+                            this._updateToken(function() {
+                                // restore the original error handler
+                                options.error = oldError;
+                                delete options.xhr;  // remove the failed (401) xhr from the last try.
+
+                                // update authorization header
+                                this._setHeaders( options );
+
+                                Backbone.sync.call(this, method, model, options);
+                            }.bind(this));
+                        } else if (oldError) {
+                            // fall back to the original error handler
+                            oldError.call(this, xhr, textStatus, errorThrown);
+                        }
+                    }.bind(this);
+                    return Backbone.sync.call(this, method, model, options);
+                },
+
+                /**
+                 * Fix up headers on an imminent AJAX sync, ensuring that the JWT token is enclosed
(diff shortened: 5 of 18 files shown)


programs/static/js/models/course_run_model.js
```diff
@@ -0,0 +1,10 @@
+define([
+        'backbone'
+    ],
+    function( Backbone ) {
+        'use strict';
+
+        return Backbone.Model.extend({
```
jimabramson: should this extend AutoAuthModel ?

jimabramson: I guess it does not need to since it has no API of its own

programs/static/js/views/course_details_view.js
```diff
@@ -45,13 +48,14 @@ define([
                     this.setDefaultModel();
                 }
 
-                //this.programModel.on('change:display_name', this.update, this);
+                this.programModel.set({cid: this.programModel.cid});
```
jimabramson: what does this do?

AlasdairSwan: I need to use ids in the DOM for the aria labels so to make them unique I am grabbing the model's cid and setting it as an attribute I can access in the underscore template.

programs/static/js/collections/course_runs_collection.js
```diff
@@ -0,0 +1,43 @@
...
+    ],
+    function( Backbone, $, apiConfig, AutoAuthCollection ) {
+        'use strict';
+
+        return AutoAuthCollection.extend({
+            allRuns: [],
+
+            initialize: function() {
+                // Ignore pagination and give me everything
+                var queries = '?org=edx&page_size=1000';
```
rlucioni: I don't think you want to hardcode `edx` as the org here. The org passed to the API should be the org code corresponding to the organization which owns a given program.

programs/static/js/models/auto_auth_model.js
```diff
@@ -1,84 +1,16 @@
...
         'backbone',
-        'jquery',
-        'js/utils/api_config'
+        'js/utils/auth_utils'
     ],
-    function( Backbone, $, apiConfig ) {
+    function( Backbone, authUtils ) {
         'use strict';
 
         return Backbone.Model.extend({
```
jimabramson: curious if it would work just to say `Backbone.Model.extend(authUtils)` here?

programs/static/js/models/api_config_model.js
```diff
@@ -6,7 +6,8 @@ define([
 
         return Backbone.Model.extend({
             defaults: {
-                baseUrl: 'http://127.0.0.1:8004/api/v1/',
+                lmsBaseUrl: 'http://127.0.0.1:8001/api/v1/',
+                programsBaseUrl: 'http://127.0.0.1:8004/api/v1/',
```
rlucioni: Given the way these base URLs are appended to paths throughout the application (e.g., `lmsBaseUrl + 'api/courses/v1/courses/'`), you may want to remove `/api/v1/` from these. That will make them true "base URLs."

programs/static/js/views/course_details_view.js
```diff
@@ -45,13 +48,14 @@ define([
...
+                this.programModel.on('change:display_name', this.update, this);
                 this.render();
             },
 
             render: function() {
                 var data = _.extend( this.programModel.toJSON(), {
-                    cid: this.programModel.cid
+                    availableRuns: this.availableCourseRuns
```
jimabramson: i think this should be `this.availableRuns` (i can't see `availableCourseRuns` being set anywhere)

programs/static/js/views/course_details_view.js
```diff
@@ -158,16 +166,15 @@ define([
...
+            },
 
-            // TODO: Determine whether this is still needed
-            // update: function() {
-            //     var data = _.extend( this.programModel.toJSON(), {
-            //         cid: this.programModel.cid
-            //     });
+            update: function() {
+                var data = _.extend( this.programModel.toJSON(), {
+                    availableRuns: this.availableCourseRuns
```
jimabramson: same comment about availableCourseRuns vs. availableRuns

programs/templates/program.html
```diff
@@ -6,7 +6,7 @@
 
 {% block content %}
     <!-- N.B. Authentication requires a running instance of Studio. -->
-    <div class="js-program-admin program-app layout-1q3q layout-reversed" data-home-url="/program" data-api-url="/api/v1/" data-auth-url="http://127.0.0.1:8001/programs/id_token/"></div>
+    <div class="js-program-admin program-app layout-1q3q layout-reversed" data-home-url="/program" data-lms-api-url="http://127.0.0.1:8000/" data-programs-api-url="/api/v1/" data-auth-url="http://127.0.0.1:8001/programs/id_token/"></div>
```
jimabramson: don't forget to remove hard-coded URLs

rlucioni: In this case, I think these hardcoded URLs should be kept. The application pulls these URLs from the DOM and sets them on the apiConfig object.

programs/static/js/models/course_run_model.js
rlucioni: Since nothing's being passed to `extend()` here, is it sufficient to return `Backbone.Model`?

programs/static/js/views/program_details_view.js
```diff
@@ -58,7 +63,8 @@ define([
             addCourse: function() {
                 return new CourseView({
                     collection: this.model.get('course_codes'),
-                    courseList: this.courseList
+                    courseList: this.courseList,
```
rlucioni: Is the `CourseListModel ` model still necessary?

programs/static/js/views/program_details_view.js
```diff
@@ -150,6 +156,11 @@ define([
                 this.model.patch( data );
             },
 
+            setAvailableCourseRuns: function() {
+                // TODO: Once API is saving data update to remove any runs already saved
+                return true;
```
rlucioni: Which API are you referring to here? Is the idea to use this method to patch the program as new course codes and runs are selected?

AlasdairSwan: That is indeed the idea @rlucioni, which is the next (and final) sub-task for this story.

programs/static/js/models/api_config_model.js
AlasdairSwan: Good point, I will update

programs/static/js/utils/auth_utils.js
```diff
@@ -0,0 +1,83 @@
+define([
+        'jquery',
+        'underscore',
+        'js/utils/api_config'
+    ],
+    function( $, _, apiConfig ) {
+        'use strict';
+
+        return {
+            auto: {
```
jimabramson: why does this have to be nested?

programs/static/js/models/course_model.js
```diff
@@ -19,6 +19,13 @@ define([
                     required: true,
                     maxLength: 128
                 }
+            },
+
+            defaults: {
+                display_name: false,
+                key: false,
+                organizations: [],
```
jimabramson: only one organization can be associated with a course

programs/static/js/utils/auth_utils.js
AlasdairSwan: this is specific to auto authentication so I was keeping the option of adding other auth utils to this file

AlasdairSwan: And enabling the ability to just add ```authUtils.auto``` into the collection/model extend

programs/static/js/views/course_run_view.js
```diff
@@ -44,20 +53,43 @@ define([
...
+                updatedRuns = _.filter( runs, function( obj ) {
+                    return obj.start !== startDate &&
+                           obj.course_key !== courseKey;
+                });
+
+                this.courseModel.set({
+                    course_runs: updatedRuns
                 });
 
+                this.courseRuns.addRun(courseKey);
```
jimabramson: assuming `this.courseRuns` holds the superset of possible selections, could the variable be renamed `this.courseRunChoices` or similar?

programs/static/js/collections/course_runs_collection.js
```diff
@@ -0,0 +1,51 @@
...
+            // Adds a run back into the model for selection
+            addRun: function(id) {
+                var courseRun = _.findWhere( this.allRuns, { course_id: id });
+
+                this.create(courseRun);
+            },
+
+            getOrganizationsStr: function(orgs) {
+                var names = _.pluck(orgs, 'display_name'),
+                    str = names.join('&org=');
```
jimabramson: as mentioned below i don't think there's any case for working with multiple orgs for a single course, so this can be simplified.

programs/static/js/views/course_details_view.js
```diff
@@ -94,13 +113,24 @@ define([
...
-                var runs = this.courseRuns;
+                var runs = this.courseRunViews;
 
                 _.each( runs, function( run ) {
-                    run.destroy();
+                    run.removeRun();
                 });
             },
 
+            // Format datat to be passed to the template
```
jimabramson: nitpicky: typo

programs/static/js/utils/auth_utils.js
jimabramson: understood.  Maybe consider renaming it to something like AutoAuthSync or similar, so it's more clear that it defines a base object type that things will inherit from.

programs/static/js/views/course_run_view.js
AlasdairSwan: Sure, that works.

PR #66, opened 2015-12-14
Eagerly load serialized data to prevent cascades of performance-degrading queries

rlucioni:
Prevents serialized nested relationships from incurring n+1 round-trips to the database. ECOM-3173.

@jimabramson please review. I've not yet been able to figure out the test failures. The Django debug toolbar indicates that prefetching cuts the number of queries made on GETs to /api/v1/programs/ by about 70% (e.g., with 6 programs, 3 orgs, 5 course codes, and 9 course run modes, the query count drops from 39 to 12).

### DIFF
--- a/programs/apps/api/v1/views.py
+++ b/programs/apps/api/v1/views.py
@@ -1,6 +1,7 @@
 """
 Programs API views (v1).
 """
+from django.db.models import Prefetch
 from rest_framework import mixins, viewsets, parsers as drf_parsers
 
 from programs.apps.programs import models
@@ -56,7 +57,6 @@ class ProgramsViewSet(
 
     """
     permission_classes = (permissions.IsAdminGroupOrReadOnly, )
-    queryset = models.Program.objects.all()
     filter_backends = (
         filters.ProgramStatusRoleFilterBackend,
         filters.ProgramStatusQueryFilterBackend,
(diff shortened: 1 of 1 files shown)


rlucioni:
I was able to shave off an additional 3 queries by using Django's `Prefetch()` objects. The total number of queries in my example has now dropped to 9. Only 5 of these are used to retrieve Programs data, as opposed to 35 without prefetching.

jimabramson:
@rlucioni a clue: in a debugging session just now I saw that the serializers are still doing the right thing with respect to nested writes; however, the `.data` attribute of the top-level program serializer does not get updated to reflect the new state of the database.  To confirm this you can disable the assertions in the failing tests which check the response data output (coming from `ProgramSerializer.data`) and leave the checks enabled that look at the db contents (I did this `ProgramsViewTests.test_update_course_codes` and saw the test succeed).

jimabramson:
ding! https://github.com/tomchristie/django-rest-framework/issues/2442

rlucioni:
Nice find @jimabramson! I'll update `get_queryset()` to only prefetch on GET requests.

programs/apps/api/v1/views.py
```diff
@@ -65,6 +65,29 @@ class ProgramsViewSet(
     serializer_class = serializers.ProgramSerializer
     parser_classes = (edx_parsers.MergePatchParser, drf_parsers.JSONParser)
 
+    def get_queryset(self):
+        """Perform eager loading of data to prevent a cascade of performance-degrading queries."""
+        queryset = models.Program.objects.all()
```
jimabramson: nitpick: variable not used when request.method != 'GET'

jimabramson:
:+1:  thanks for taking this one on @rlucioni

programs/apps/api/v1/views.py
rlucioni: Which variable are you referring to? `queryset` is returned unmodified when `self.request.method != 'GET'`.

jimabramson: I just meant that it could be inlined with the return statement instead of
making a variable for it.  Feel free to ignore.

PR #69, opened 2015-12-16
Use RequireJS Optimizer to minify JS and CSS

rlucioni:
Yields minified files with unhashed names we can use to fetch these assets from Studio. ECOM-2862.

### DIFF
--- a/build.js
+++ b/build.js
@@ -1,20 +1,44 @@
 /* jshint asi:true, expr:true */
 ({
+    // For documentation of the directives in this build file, refer to
+    // the example require.js build file located at:
+    // https://github.com/jrburke/r.js/blob/master/build/example.build.js
+
+    // See: https://github.com/jrburke/r.js/blob/master/build/example.build.js#L27-L37
     mainConfigFile: 'programs/static/js/config.js',
+
+    // See: https://github.com/jrburke/r.js/blob/master/build/example.build.js#L21-L25
     baseUrl: 'programs/static',
+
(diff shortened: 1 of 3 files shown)


build.js
```diff
@@ -1,20 +1,44 @@
...
     wrapShim: true,
 
-    // Disable all optimization. django-compressor will handle that for us.
-    optimizeCss: false,
-    optimize: 'none',
+    // See: https://github.com/jrburke/r.js/blob/master/build/example.build.js#L208-L220
+    optimizeCss: 'standard',
+
+    // See: https://github.com/jrburke/r.js/blob/master/build/example.build.js#L99-L110
+    optimize: 'uglify2',
```
feanil: How does uglify get installed?

feanil:
If your assets are un-hashed this will prevent future use of CDNs with programs.  Is there a different part of the path that is going to get hashed or is this a known tradeoff?

build.js
rlucioni: @feanil I believe it's included with r.js. @AlasdairSwan correct me if I'm mistaken.

rlucioni:
@feanil Django Compressor is still being run, giving us hashed assets to work with if and when we need them. In the meantime, the use of unhashed assets doesn't preclude the use of cache-control headers. I'll look into extending https://github.com/edx/configuration/pull/2602 to add these headers to Programs' Nginx configuration.

AlasdairSwan:
:+1:

PR #70, opened 2015-12-17
Fix incorrect routing after new program is saved

rlucioni:
Formerly redirected to '/program/program/:id'. Now redirects to '/program/:id'. ECOM-3120.

### DIFF
--- a/programs/static/js/views/program_creator_view.js
+++ b/programs/static/js/views/program_creator_view.js
@@ -103,7 +103,7 @@ define([
             },
 
             saveSuccess: function() {
-                this.goToView( 'program/' + this.model.get( 'id' ) );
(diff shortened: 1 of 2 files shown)


jimabramson:
:+1:   so it was that simple...

programs/static/js/views/program_creator_view.js
```diff
@@ -103,7 +103,7 @@ define([
             },
 
             saveSuccess: function() {
-                this.goToView( 'program/' + this.model.get( 'id' ) );
+                this.goToView( String( this.model.get( 'id' ) ) );
```
AlasdairSwan: I don't think you need to wrap it in ```String()```, was it not working with just returning the id?

AlasdairSwan:
Great, ship it :+1:

PR #71, opened 2015-12-17
Reorganize dependencies to allow gulp tasks to be run during production deployments

rlucioni:
This is in response to an issue causing CI builds to fail in Jenkins, introduced by https://github.com/edx/configuration/pull/2602. The failures are due to the fact that gulp and its dependencies weren't included in Programs' production node dependencies.

@clintonb @jibsheet please review. I've manually started [a build](http://jenkins.edx.org:8080/job/ansible-provision-continuous-integration/21146/) pointing at this branch to verify that these changes fix the problem.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,16 +7,16 @@
   },
   "dependencies": {
     "bower": "^1.4.1",
-    "requirejs": "^2.1.18"
-  },
-  "devDependencies": {
+    "requirejs": "^2.1.18",
     "del": "*",
     "gulp": "^3.9.0",
     "gulp-jscs": "^3.0.1",
     "gulp-jshint": "^1.11.2",
     "gulp-sass": "*",
+    "karma": "^0.13.2"
+  },
+  "devDependencies": {
     "jasmine-core": "^2.3.4",
-    "karma": "^0.13.2",
     "karma-coverage": "^0.5.2",
     "karma-firefox-launcher": "^0.1.6",
     "karma-jasmine": "^0.3.6",


rlucioni:
The build continues to fail, though with a new error:
```
22:52:22 Error: The `libsass` binding was not found in /edx/app/programs/programs/node_modules/node-sass/vendor/linux-x64-47/binding.node
22:52:22 This usually happens because your node version has changed.
22:52:22 Run `npm rebuild node-sass` to build the binding for your current node version.
```

I'm investigating.

clintonb:
:+1:

rlucioni:
According to https://github.com/sass/node-sass/issues/1162#issuecomment-141852021, this error occurs when `npm install` was run with a different version of `node` than the version running at the time Sass compilation is attempted.

It took a while, but I noticed that the version of `npm` installed by `nodeenv` is not the latest supported version. I'm running [another build](http://jenkins.edx.org:8080/job/ansible-provision-continuous-integration/21154/) which will update `npm` before installing node dependencies.

rlucioni:
@jibsheet I'm still working on this. CI now appears to be failing due to an unrelated issue when attempting to [install the mongo server and recommends](http://jenkins.edx.org:8080/job/ansible-provision-continuous-integration/21160/).

The changes in this PR are a step in the right direction, so I'm going to go ahead and merge. Any other changes will most likely affect configuration.

jibsheet:
I think you still need this, right?  If so, :+1: 
Thanks for digging into this, I believe the mongo errors you ran into later were timeouts from the server hosted by mongodb.org (they released a big release and were under load serving updated packages today).

PR #73, opened 2015-12-21
Correct handling of optional, empty fields

rlucioni:
Avoids use of Field.null on string-based fields, as suggested by [Django's documentation](https://docs.djangoproject.com/en/1.9/ref/models/fields/#null). Enforces uniqueness of non-empty marketing slugs in code instead of at the database level. ECOM-2918.

The included migration will replace any NULL values in the affected fields with the empty string. The migration will lock affected tables when it is run, but we should be able to tolerate this. Presently, users only interact with the Programs service via the Programs API. The LMS caches the result of successful calls to the Programs API and is robust to failed calls. Worst case, a handful of users may not see XSeries messaging on their dashboard for a brief period while the migration is applied.

@jimabramson please review when you're able. @AlasdairSwan FYI.

### DIFF
--- a/programs/apps/programs/models.py
+++ b/programs/apps/programs/models.py
@@ -32,7 +32,6 @@ class Program(TimeStampedModel):
     subtitle = models.CharField(
         help_text=_('A brief, descriptive subtitle for the Program.'),
         max_length=255,
-        null=True,
         blank=True,
     )
 
@@ -60,23 +59,20 @@ class Program(TimeStampedModel):
 
     marketing_slug = models.CharField(
         help_text=_('Slug used to generate links to the marketing site'),
-        unique=True,
-        null=True,
         blank=True,
         max_length=255
     )
 
     def save(self, *a, **kw):
-        """Validate that the marketing slug is not null if program category is
-        xseries and program status is active.
         """
-        # Checks if the program category is xseries and program status is active.
+        Verify that the marketing slug is not empty if the user has attempted
+        to activate an XSeries.
+        """
         if self.category == constants.ProgramCategory.XSERIES and self.status == constants.ProgramStatus.ACTIVE:
-            # Validate that marketing slug is not null.
             if not self.marketing_slug:
-                raise ValidationError(
-                    _("Active XSeries Programs must have a valid marketing slug.")
-                )
+                raise ValidationError(_(
+                    "Active XSeries Programs must have a valid marketing slug."
+                ))
 
         return super(Program, self).save(*a, **kw)
 
@@ -129,9 +125,10 @@ class ProgramOrganization(TimeStampedModel):
         This is a temporary, application-level constraint.
         """
         if not self.id:  # pylint: disable=no-member
-            # before creating, check program is not already associated with some organization
+            # Before creating, check program is not already associated with some organization
             if ProgramOrganization.objects.filter(program=self.program).exists():
                 raise ValidationError(_('Cannot associate multiple organizations with a program.'))
+
         return super(ProgramOrganization, self).save(*a, **kw)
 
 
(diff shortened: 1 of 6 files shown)


programs/apps/programs/models.py
```diff
@@ -60,23 +59,30 @@ class Program(TimeStampedModel):
...
+        # Enforce uniqueness. String-based fields such as CharField and TextField store empty string values
+        # as empty strings, not as NULL. As such, we want to avoid enforcing uniqueness at the database level,
+        # since only one row would be allowed the empty string.
+        # See: https://docs.djangoproject.com/en/1.9/ref/models/fields/#null.
+        if self.marketing_slug and Program.objects.filter(
+                marketing_slug=self.marketing_slug
+        ).exclude(id=self.id).exists():  # pylint: disable=no-member
+            raise ValidationError(_(
+                "A Program with the provided marketing slug already exists. The marketing slug must be unique."
+            ))
```
jimabramson: @rlucioni not sure if we discussed this once before but maybe it's simpler just to lift the uniqueness restriction on marketing slug ?  in retrospect i don't think it's really providing much benefit, let alone does it seem worth writing custom code to enforce it.

programs/apps/programs/models.py
```diff
@@ -237,20 +242,20 @@ class Meta(object):  # pylint: disable=missing-docstring
 
     def save(self, *_a, **__kw):
         """
-        The database will not enforce a unique constraint over NULL values in
+        The database will not enforce a unique constraint over empty values in
```
jimabramson: @rlucioni now that NULLs are being excised I think this custom handling can just be removed; I'd expect the empty string would be treated as a unique value for the sku column like any other string value.

programs/apps/programs/tests/test_models.py
```diff
@@ -197,14 +183,14 @@ def setUp(self):
     def test_unique_null_sku(self):
```
jimabramson: @rlucioni if you agree with my suggestion to remove the manual uniqueness validation on sku above (and this test still passes after the change, confirming it's not needed), then let's remove the test.

jimabramson:
@rlucioni reviewed and suggested some simplifications.

Thanks for thinking through the migration strategy.  I can imagine ways of deploying this in two steps to achieve zero downtime but given how well-insulated we are, I don't think it's worth complicating things.

programs/apps/programs/models.py
rlucioni: @jimabramson agreed. I'm all for less code where possible, so I'll remove this.

programs/apps/programs/models.py
rlucioni: @jimabramson we need to keep some of this custom handling, don't we? We don't want to allow a course code to have duplicate course run modes (e.g., two course run modes with the same course_key and mode_slug for the same course code).

I've pushed up a commit which removes manual validation of SKU uniqueness but continues to enforce program_course_code, course_key and mode_slug uniqueness. Doing so required modifying the test you tagged below. Let me know what you think.

rlucioni: Do'h. I totally missed that the unique_together will handle this validation for us. Give me a moment to clean this up.

rlucioni: Turns out that the unique_together constraint only gives us free validation when using Django's ModelForm. Attempts to directly save a model object violating this constraint will raise an IntegrityError. That being the case, I think it's in our interest to keep this manual validation around.

rlucioni:
@jimabramson I've addressed your comments. I'd appreciate it if you could take another look, when you're able.

programs/apps/programs/models.py
jimabramson: alright.  thanks for trying that out.

jimabramson:
:+1:

PR #74, opened 2015-12-22
ECOM-3046 saving courses and runs to program model

AlasdairSwan:


programs/static/js/models/course_run_model.js
```diff
@@ -4,6 +4,14 @@ define([
     function( Backbone ) {
         'use strict';
 
-        return Backbone.Model.extend({});
+        return Backbone.Model.extend({
+        	defaults: {
```
rlucioni: Minor: are there tabs in here?

rlucioni: If so, I'd replace them with our standard four spaces.

programs/static/js/views/course_run_view.js
```diff
@@ -73,11 +88,15 @@ define([
...
-                    runs = this.courseModel.get('run_modes');
-
-                data.course_key = id;
-                this.model.set(data);
+                    runObj = _.findWhere(this.courseRuns.allRuns, {course_id: id}),
+                    /**
+                     *  NB: cloning the array so the model will fire a change event when
+                     *  the updated version is saved back to the model
+                     */
+                    runs = _.clone(this.courseModel.get('run_modes')),
```
rlucioni: When is `runs` updated? I see it saved back to the model on line 101, but I don't see it modified at any point after being copied.

AlasdairSwan: updatedRuns is a filtered version of runs. Let me double check whether the clone is needed here, it may not be but then with the weirdness around models listening to changes in arrays it very well may be still needed. Will let you know.

rlucioni: I somehow commented here instead of on line 96.

rlucioni: I meant to post my question here. This copy of `run_modes` doesn't seem to be filtered.

AlasdairSwan: So was this comment intended for line 96? If so runs.push on line 100 updates runs.

rlucioni: It was, I see it. Thank you.

programs/static/js/views/course_run_view.js
```diff
@@ -50,16 +51,30 @@ define([
                 this.remove();
             },
 
+            // Data returned from courseList API is not the correct format
+            formatData: function( data ) {
+                return {
+                    course_id: data.course_id,
+                    course_key: data.course_id,
```
rlucioni: Is `course_id` required here? The course run model has a `course_key` field, but not `course_id` field.

rlucioni:
@AlasdairSwan I have a question about data formatting, but this is looking good to me otherwise.

programs/static/js/views/course_run_view.js
AlasdairSwan: runs.push on line 100 updates runs.

programs/static/js/views/course_run_view.js
AlasdairSwan: I think I added it to make it easier to reference the allRuns data, but I should be able to update to not need to.

rlucioni:
:+1:

PR #75, opened 2015-12-30
Alasdair/ecom 3047 embedded app styling

AlasdairSwan:
@rlucioni the issue is that the unit being used for fonts, padding and margins in the UX Pattern Library (UXPL) is relative to the font-size set on the html element. In studio this is set to 62.5%, however the UXPL assumes it is set to 100%. Because the page in studio has other elements that are native to studio that need to be styled we cannot adjust the html element's font-size on this page so I have added a Sass partial that sets px values for the elements to resolve the issue. In the long-term the UXPL will be added to studio and we will be able to remove this partial.

### DIFF
--- a/programs/static/sass/_app-container.scss
+++ b/programs/static/sass/_app-container.scss
@@ -1,0 +1,10 @@
+
+// ------------------------------
+// Programs: App Container
+
+// About: styling for setting up the wrapper.
+.program-app {
+  &.layout-1q3q {
+    max-width: 1250px;
+  }
+}
--- a/programs/static/sass/_app-typography.scss
+++ b/programs/static/sass/_app-typography.scss
@@ -1,0 +1,79 @@
+
+// ------------------------------
+// Programs: App Fonts
+
+// About: resolving issue with rem units when html font-size is not 100%.
+// the issue is that the unit being used for fonts, padding and margins in the
+// UX Pattern Library (UXPL) is relative to the font-size set on the html element.
+// In studio this is set to 62.5%, however the UXPL assumes it is set to 100%. Because
+// the page in studio has other elements that are native to studio that need to be styled
+// we cannot adjust the html element's font-size on this page so I have added a Sass partial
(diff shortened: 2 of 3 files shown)


AlasdairSwan:
screen shot 2015-12-30 at 10 02 41 am

programs/static/sass/_app-typography.scss
```diff
@@ -0,0 +1,69 @@
+
+// ------------------------------
+// Programs: App Fonts
+
+// About: resolving issue with rem units when html font-size is not 100%.
+.program-app {
+	.hd-3 {
```
rlucioni: Minor: this line is indented too far, and I see tab characters throughout. Can you clean that up?

programs/static/sass/_app-typography.scss
```diff
@@ -0,0 +1,69 @@
+
+// ------------------------------
+// Programs: App Fonts
+
+// About: resolving issue with rem units when html font-size is not 100%.
```
rlucioni: Can you extend this comment with context from the first comment on this PR? I think it's worth noting that this could be removed once the UXPL is incorporated into Studio.

AlasdairSwan:
screen shot 2015-12-30 at 10 20 44 am

programs/static/sass/_app-container.scss
```diff
@@ -0,0 +1,10 @@
+
+// ------------------------------
+// Programs: App Container
+
+// About: styling for setting up the wrapper.
+.program-app {
+    &.layout-1q3q {
+        max-width: 1250px;
```
rlucioni: For my own edification, why use px instead of % here?

AlasdairSwan: I am getting it to line up with studio. The max-width is usually set as a px value 'tho, with the width being a percentage.

rlucioni: Got it, you're constraining the size of the element to the given px value.

rlucioni:
:+1: once commits merged.

PR #76, opened 2015-12-30
ECOM-3047 updated to use programs api url instead of programs base url

AlasdairSwan:
@rlucioni the changes needed in light of https://github.com/edx/edx-platform/pull/11101


rlucioni:
:+1: once tests pass.

PR #77, opened 2016-01-04
ECOM-3319 updated so only loading font files in local app.

AlasdairSwan:
Assumption is being made that other apps using the programs IDA will already have loaded OpenSans and edx-icons fonts
@jimabramson @rlucioni please review
**NB** there will be an accompanying PR to add the font file to studio.

### DIFF
--- a/programs/static/sass/_scoped-pattern-library.scss
+++ b/programs/static/sass/_scoped-pattern-library.scss
@@ -16,7 +16,9 @@
     @import '../bower_components/edx-pattern-library/pattern-library/sass/utilities/variables-ltr';
 }
 
-@import '../bower_components/edx-pattern-library/pattern-library/sass/utilities/variables';
+// Manually copied version of pattern library's variables to remove fonts
+@import 'uxpl-variables-sans-fonts';
+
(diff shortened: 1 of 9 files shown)


programs/static/sass/_scoped-pattern-library.scss
```diff
@@ -16,7 +16,9 @@
     @import '../bower_components/edx-pattern-library/pattern-library/sass/utilities/variables-ltr';
 }
 
-@import '../bower_components/edx-pattern-library/pattern-library/sass/utilities/variables';
+// Manually copied version of pattern library's variables to remove fonts
```
jimabramson: this seems suboptimal from a maintenance perspective.  are there no alternatives?

AlasdairSwan: Agreed, but unfortunately there are no alternatives due to what we are trying to do. The Pattern Library was not developed to be scoped to a single section of a page (so it was co-exist with the styles already added to said page). As a result our implementation of it has become pretty customized. The ultimate goal is for studio to already have the Pattern Library loaded, in which case we can rip all of this out and use a nice, clean version. Until then this sort of hack is necessary.

jimabramson:
LGTM :+1:

PR #78, opened 2016-01-07
Add service user for credentials

ahsan-ul-haq:
ECOM-3382

@zubair-arbi @awais786 @tasawernawaz

### DIFF
--- a/programs/apps/core/migrations/0002_add_credentials_service_user.py
+++ b/programs/apps/core/migrations/0002_add_credentials_service_user.py
@@ -1,0 +1,33 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+from django.conf import settings
+from programs.apps.core.models import User
+
+
+def add_service_user(apps, schema_editor):
+    """Add service user."""
+    user = User.objects.create(username=settings.CREDENTIALS_SERVICE_USERNAME)
+    user.set_unusable_password()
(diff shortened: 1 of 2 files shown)


zubair-arbi:
:+1:

ahsan-ul-haq:
@jimabramson please review
FYI: @rlucioni

programs/apps/core/migrations/0002_add_credentials_service_user.py
```diff
@@ -0,0 +1,30 @@
...
+def add_service_user(apps, schema_editor):
+    """Add service user."""
+    user = User.objects.create(username=settings.CREDENTIALS_SERVICE_USERNAME)
+    user.set_unusable_password()
+    user.save()
+
+
+def remove_service_user(apps, schema_editor):
+    """Remove service user."""
+    User.objects.get(username=settings.CREDENTIALS_SERVICE_USERNAME).delete()
```
jimabramson: i think i'd add a try/except around this, so that if the user account doesn't exist for some reason, it won't cause an incomplete / failed migration.

jimabramson:
@ahsan-ul-haq this looks good :+1:  - please be aware however that we'll soon have a new way of doing this.  when https://github.com/edx/edx-django-extensions/pull/1 lands, we'll be able to replace the hardcoded service user (and this migration) with scripts in configuration to ensure the user exists.  At that point we will be able to remove this migration.

PR #79, opened 2016-01-07
get username from DOM, fixing course API requests.

jimabramson:
@rlucioni @AlasdairSwan FYI.

Also fixes the query string for org filtering.

### DIFF
--- a/programs/templates/program.html
+++ b/programs/templates/program.html
@@ -5,8 +5,8 @@
 {% load staticfiles %}
 
 {% block content %}
-    <!-- N.B. Authentication requires a running instance of Studio. -->
-    <div id="content" class="js-program-admin program-app layout-1q3q layout-reversed" data-home-url="/program" data-lms-base-url="http://127.0.0.1:8000/" data-programs-api-url="http://127.0.0.1:8004/api/v1/" data-auth-url="http://127.0.0.1:8001/programs/id_token/"></div>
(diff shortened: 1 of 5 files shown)


programs/templates/program.html
```diff
@@ -5,8 +5,8 @@
 {% load staticfiles %}
 
 {% block content %}
-    <!-- N.B. Authentication requires a running instance of Studio. -->
-    <div id="content" class="js-program-admin program-app layout-1q3q layout-reversed" data-home-url="/program" data-lms-base-url="http://127.0.0.1:8000/" data-programs-api-url="http://127.0.0.1:8004/api/v1/" data-auth-url="http://127.0.0.1:8001/programs/id_token/"></div>
+    <!-- N.B. Authentication requires a running instance of Studio, and requires the use of user account "staff". -->
+    <div id="content" class="js-program-admin program-app layout-1q3q layout-reversed" data-home-url="/program" data-lms-base-url="http://127.0.0.1:8000/" data-programs-api-url="http://127.0.0.1:8004/api/v1/" data-auth-url="http://127.0.0.1:8001/programs/id_token/" data-username="staff"></div>
```
jimabramson: This is a little unfortunate, but really not much more problematic than the other defaults we have hardcoded here.

jimabramson:
btw depends on https://github.com/edx/edx-platform/pull/11170

rlucioni:
:+1: assuming it's not possible to get the username off the request.

PR #80, opened 2016-01-07
wait to set organization model urlRoot until init

jimabramson:
@AlasdairSwan note this is a simpler fix than you had suggested but it seems to work out.  LMK what you think.


rlucioni:
:+1:

AlasdairSwan:
I like it @jimabramson :+1:

PR #81, opened 2016-01-07
suppress Studio’s error popup on the initial 401

jimabramson:
@rlucioni @AlasdairSwan FYI, yet another little fix.

This avoids 'Studio is having trouble saving your work' from popping up after a 401 that will be retried using the auth_utils.  This is not a jQuery API but something that's already implemented in Studio, see https://github.com/edx/edx-platform/blob/master/cms/static/coffee/src/main.coffee#L22

There is a theoretical corner case in which this approach may cause a legitimate error NOT to trigger the popup, however I haven't been able to create such a situation on purpose, and I'm frankly not too worried about it.

### DIFF
--- a/programs/static/js/utils/auth_utils.js
+++ b/programs/static/js/utils/auth_utils.js
@@ -26,12 +26,15 @@ define([
 
                     this._setHeaders( options );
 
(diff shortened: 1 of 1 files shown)


rlucioni:
:+1: Thank you very much for fixing this.

AlasdairSwan:
:+1:

PR #82, opened 2016-01-08
substitute _.findIndex to fix course run selection

jimabramson:
@AlasdairSwan sanity check?


AlasdairSwan:
Looks good :+1:  - Andy already had updating underscore on his roadmap so that's good.

PR #85, opened 2016-01-11
ECOM-3404 updated course_runs_colleciton to use a custom createNew fu…

AlasdairSwan:
…nction instead of the standard create to prevent attempts to save the collection after updates are made
@jimabramson please review. The fix was a lot simpler than first feared.

### DIFF
--- a/programs/static/js/collections/course_runs_collection.js
+++ b/programs/static/js/collections/course_runs_collection.js
@@ -19,6 +19,21 @@ define([
                 this.url = apiConfig.get('lmsBaseUrl') + 'api/courses/v1/courses/' + queries;
             },
 
+            /*
+             *  Abridged version of Backbone.Collection.Create that does not
+             *  save the updated Collection back to the server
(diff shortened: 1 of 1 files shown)


programs/static/js/collections/course_runs_collection.js
```diff
@@ -19,6 +19,21 @@ define([
                 this.url = apiConfig.get('lmsBaseUrl') + 'api/courses/v1/courses/' + queries;
             },
 
+            /*
+             *  Abridged version of Backbone.Collection.Create that does not
+             *  save the updated Collection back to the server
+             *  (code based on original function - http://backbonejs.org/docs/backbone.html#section-134)
+             */
+            createNew: function(model, options) {
```
jimabramson: does this have to be named `createNew` instead of overriding `create` ?

AlasdairSwan: I was trying to make it explicit that it's not the exact same as create, but yeah I will update to be called create because the whole point of Backbone is extendability.

jimabramson:
:+1:

PR #86, opened 2016-01-11
don’t complain about empty sku in course run mode

jimabramson:
@AlasdairSwan fixes problems with course runs that don't have any SKU


AlasdairSwan:
Once rebase and tests pass :+1:

PR #87, opened 2016-01-11
Order orgs by key in dropdown

jimabramson:
@AlasdairSwan makes the org dropdown easier to use - the order was previously arbitrary


AlasdairSwan:
:+1:

PR #88, opened 2016-01-13
wait until initialize() to read from apiConfig.

jimabramson:
Fixes an issue I encountered on the sandbox, and might otherwise have bit on prod.  If the apiConfig is accessed at load time (while resolving dependencies) its default values may not yet have been overwritten by the app.

@AlasdairSwan @rlucioni FYI.


rlucioni:
:+1: nice catch.

AlasdairSwan:
:+1:

PR #89, opened 2016-01-21
filter course runs by org key, not display_name

jimabramson:
@AlasdairSwan @rlucioni I'm not sure how this slipped through, but the query filter we've been using was incorrect.  Please review.


rlucioni:
@jimabramson I'm surprised we didn't catch that. :+1:

jimabramson:
@rlucioni I think it's because typically we test with "edX" whose display name and key are the same!

rlucioni:
@jimabramson good point!

PR #90, opened 2016-01-22
added script to load orgs manually

jimabramson:
@rlucioni FYI.  I don't think this will be used more than once or twice so I haven't put a lot of effort into it.  I am not even that concerned with merging this branch, but I wanted to make sure this script exists somewhere other than my own hard drive :-)

### DIFF
--- a/scripts/load_organizations.py
+++ b/scripts/load_organizations.py
@@ -1,0 +1,49 @@
+"""
+This utility allows the Organization model to be populated by generating
+POST requests to the organizations API endpoint using a csv input file.  (It
+is not very polished, because it is not intended for long-term or frequent
+use.)
+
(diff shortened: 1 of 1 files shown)


rlucioni:
@jimabramson seeing as how it's a helpful tool to have when bootstrapping an instance of Programs, it might be worth merging (along with some documentation).

schenedx:
@rlucioni Could you shepherd this PR to merge? I think we should merge this script because it might be useful in the long run.

rlucioni:
@schenedx I'm closing this. It's superseded by the [sync_orgs](https://github.com/edx/programs/blob/master/programs/apps/programs/management/commands/sync_orgs.py#L19) management command.

PR #91, opened 2016-01-22
Add interface for program completion endpoint

rlucioni:
The endpoint currently returns all program IDs, but will be refined soon. ECOM-3352.

@jimabramson

### DIFF
--- a/programs/apps/api/v1/tests/test_views.py
+++ b/programs/apps/api/v1/tests/test_views.py
@@ -45,22 +45,25 @@ class ProgramsViewTests(JwtMixin, TestCase):
         instance = ProgramFactory.build(**kwargs)
         return {k: getattr(instance, k) for k in POST_FIELDS}
 
-    def _make_request(self, method='get', program_id=None, data=None, admin=False):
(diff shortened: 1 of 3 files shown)


jimabramson:
@rlucioni looks good, although we need to update the request payload as mentioned in https://github.com/edx/edx-platform/pull/11318/files#r50644758

Curious why you chose PUT over POST for this.

rlucioni:
@jimabramson I went with PUT because it's meant to be idempotent, unlike POST. Requests to this endpoint should be idempotent.

rlucioni:
@jimabramson request payload updated to reflect changes in https://github.com/edx/edx-platform/pull/11318/.

rlucioni:
@jimabramson updated to use POST in light of our discussion.

jimabramson:
:+1: once tests pass.   Note to selves: update API documentation.

rlucioni:
@jimabramson when you say API documentation, are you referring to the viewset docstring?

jimabramson:
@rlucioni yeah, and whatever else exists; i notice there's no doc updates introduced in this PR but I consider that OK as the endpoint isn't actually implemented yet

PR #92, opened 2016-01-26
Implement program completion logic

rlucioni:
Completes the program completion endpoint. ECOM-3353.

@jimabramson please review.

### DIFF
--- a/programs/apps/api/v1/tests/test_views.py
+++ b/programs/apps/api/v1/tests/test_views.py
@@ -2,6 +2,7 @@
 Tests for Programs API views (v1).
 """
 import datetime
+import itertools
 import json
 
 import ddt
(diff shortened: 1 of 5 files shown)


jimabramson:
@peter-fogg please review also

jimabramson:
travis failure is quality related

rlucioni:
@jimabramson quality violations fixed.

programs/apps/api/v1/views.py
```diff
@@ -52,7 +53,25 @@ class ProgramsViewSet(
...
+        {'course_certs': [
+            {'course_id': 'foo', 'mode': 'bar'},
+            ...
+            {'course_id': 'baz', 'mode': 'qux'}
+        ]}
+
+        If the request is successful, the HTTP status will be 200 and the response body will
+        include JSON containing the IDs of the completed programs.
+
+        Only users with global administrative rights may find completed programs. POST requests from non-
```
jimabramson: @rlucioni didn't we decide not to restrict access in this way?

jimabramson: the permission class set on L#122 also suggests this is open to any auth user

rlucioni: Whoops, forgot to update the docstring. Updated.

programs/apps/programs/utils.py
```diff
@@ -0,0 +1,49 @@
...
+                    break
+            else:
+                is_complete = False
+
+            if is_complete:
+                continue
+            else:
+                break
+
+        if is_complete:
```
peter-fogg: I'm finding this entire method hard to parse through due to the nested looping, `break`/`continue`s, and updates of `is_complete`. Is it possible to refactor this at all to abstract away some of the control flow logic?

For example, the inner loop should be able to be reworked to something along the lines of
```python
is_complete = any((serialize_run_mode(run_mode) in complete_run_modes for run_mode in run_modes))
```
And you may be able to do something similar with the control flow of the other loops.

rlucioni: @peter-fogg nice, thanks for the suggestion. I'll look into cleaning this up.

peter-fogg: Cool; in general I like to avoid explicit loops/recursion when possible -- I suspect that there aren't really that many reasons to iterate over a collection, and most are captured by a few common combinators.

rlucioni:
@peter-fogg refactored. What do you think?

peter-fogg:
@rlucioni Nice, I like the new logic. :+1:

programs/apps/api/v1/views.py
```diff
@@ -52,9 +53,24 @@ class ProgramsViewSet(
...
+        Only users with global administrative rights may update programs. PATCH requests from non-
         admins will result in status 403.
 
+        # Find completed programs.
+        POST /api/v1/programs/complete/
+
+        Request body must include a serialized representation of completed course runs, including
+        the modes in which the runs were completed. Expected to be of the form:
+
+        {'course_certs': [
```
jimabramson: any objection to renaming this 'completed_courses' ?   it more closely matches the code and seems a little more to-the-point

rlucioni: No objections, let's do it. I was using this because it's the last key name I saw from your end.

programs/apps/api/v1/tests/test_views.py
```diff
@@ -772,26 +773,49 @@ def test_create_duplicated_name(self):
...
-    def test_program_completion_admin_required(self):
-        """Verify that the view is not accessible to non-admins."""
-        response = self._make_request(method='post', complete=True)
-        self.assertEqual(response.status_code, 403)
+        program_ids = response.data['program_ids']
+        if status == ProgramStatus.DELETED:
+            # No one should be able to complete a deleted program.
+            self.assertEqual(program_ids, [])
+        elif status == ProgramStatus.UNPUBLISHED:
+            # Only admin users should be able to complete unpublished programs.
```
jimabramson: i'm feeling a little wary of this.  the behavior arises somewhat by accident but this is a step toward codifying it despite not really having a use case.

could we instead just exclude unpublished + deleted programs from completion computations altogether?

jimabramson:
@rlucioni nice job on the checker.   a few minor concerns inline but otherwise LGTM

programs/apps/api/v1/tests/test_views.py
rlucioni: @jimabramson we can certainly exclude unpublished and deleted programs altogether. However, I wouldn't say the behavior arises by accident; it's a product of relying on the same filters we use for other API actions in this viewset. Is there a strong reason to go beyond what those filters do and introduce additional filtering logic just to strip out unpublished programs in this case?

jimabramson: you're right; accident is not the right word.  but i still think we should not be making it possible to award certificates for unpublished programs, unless you see a specific benefit in doing so.  this will lead to bogus credentials in the system.  it doesn't seem like it would take much to prevent it.

rlucioni:
@jimabramson I've addressed your comments. It is now only possible to award certificates for active and retired programs; unpublished and deleted programs are excluded from completion computations.

programs/apps/programs/utils.py
```diff
@@ -0,0 +1,51 @@
...
+        >>> completion_checker.completed_programs
+        <list of completed program IDs>
+    """
+    def __init__(self, programs, complete_run_modes):
+        self.programs = programs
+        self.complete_run_modes = complete_run_modes
+
+    @property
+    def completed_programs(self):
+        return [program.id for program in self.programs if self._is_program_complete(program)]
```
jimabramson: optional: `filter(self.programs, self._is_program_complete)`

jimabramson: oops never mind, it's program.id

jimabramson:
:+1:

PR #93, opened 2016-01-26
Sync up JS app with Django models to allow longer program names

rlucioni:
Fixes an issue preventing existing programs with long but technically legal names from being saved.

@jimabramson @AlasdairSwan


jimabramson:
:+1:

PR #96, opened 2016-02-02
Add management command to sync organization data from the LMS

rlucioni:
This command will be run on a schedule to keep data in programs up to date. I will address authentication to the LMS in a separate PR. ECOM-2202.


rlucioni:
@peter-fogg would you mind taking a look at this?

programs/apps/programs/management/commands/sync_orgs.py
```diff
@@ -0,0 +1,72 @@
...
+                        'key': org['short_name'],
+                        'display_name': org['name']
+                    }
+
+                    try:
+                        Organization.objects.get(**fields)
+                    except Organization.DoesNotExist:
+                        new_org_count += 1
+                        if commit:
+                            Organization.objects.create(**fields)
```
peter-fogg: It might make sense to wrap the database calls in `transaction.atomic` and then throw an exception at the end if `commit = False`, like we do [here](https://github.com/edx/ecommerce/blob/master/ecommerce/extensions/catalogue/management/commands/convert_honor_to_audit.py#L52-L68). I prefer this approach since it lets you branch on `commit` only once, which cleans up the code and makes it easier to modify safely in the future. This would allow using `Organization.objects.get_or_create` as well.

programs/apps/programs/tests/test_sync_orgs.py
```diff
@@ -0,0 +1,92 @@
...
+                    'active': True if i % 2 else False,
+                    'created': '2016-02-01T20:09:51.574090Z',
+                    'description': 'A great university.',
+                    'id': i,
+                    'logo': None,
+                    'modified': '2016-02-01T20:09:51.575053Z',
+                    'name': self.DISPLAY_NAME.format(i),
+                    'short_name': self.KEY.format(i)
+                }
+                for i in range(low_id, high_id + 1)
```
peter-fogg: Nit: I generally prefer `xrange` in Python 2, though it doesn't really matter here since the list is quite small.

peter-fogg:
@rlucioni A few small things, but nothing major. :+1:

programs/apps/programs/tests/test_sync_orgs.py
rlucioni: Same. However, I've started to use `range` where possible to ease the transition to Python 3. I'll leave this as-is.

peter-fogg: Ah, that's a good point.

programs/apps/programs/management/commands/sync_orgs.py
rlucioni: Thanks for the suggestion @peter-fogg. I've made this change.

PR #97, opened 2016-02-04
Use Course API's id field instead of the deprecated course_id

rlucioni:
Change being made due to https://github.com/edx/edx-platform/pull/11077.

@AlasdairSwan please review when able. @nasthagiri FYI.


AlasdairSwan:
Looks good to me :+1:

rlucioni:
@jimabramson FYI.

PR #99, opened 2016-02-17
Allow org sync management command to accept an access token

rlucioni:
The access token is used to make authenticated requests against the LMS Organizations API. ECOM-3349.

@jimabramson please review. @schenedx, how do you feel about reviewing this, too? I'm not sure if you've reviewed anything in the programs project before.

### DIFF
--- a/programs/apps/programs/management/commands/sync_orgs.py
+++ b/programs/apps/programs/management/commands/sync_orgs.py
@@ -12,6 +12,10 @@ from programs.apps.programs.models import Organization
 logger = logging.getLogger(__name__)
 
 
+class ForcedRollback(Exception):
+    pass
+
+
 class Command(BaseCommand):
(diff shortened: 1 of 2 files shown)


schenedx:
I reviewed and could not see anything that I need to critic. LGTM

programs/apps/programs/management/commands/sync_orgs.py
```diff
@@ -21,6 +25,14 @@ class Command(BaseCommand):
 
     def add_arguments(self, parser):
         parser.add_argument(
+            '--access_token',
```
jimabramson: if this argument is required, would it make sense to treat it as a positional argument, rather than an option?

rlucioni: @jimabramson yes. Updated.

jimabramson:
:+1:

PR #101, opened 2016-02-18
Display course modes in authoring tool.

jimabramson:
ECOM-3635

@AlasdairSwan @schenedx FYI.

### DIFF
--- a/programs/static/templates/course_run.underscore
+++ b/programs/static/templates/course_run.underscore
@@ -17,11 +17,16 @@
             </span>
         </button>
     <% } %>
-    <p class="copy copy-base subtitle"><%= interpolate(
+    <div class="copy copy-base subtitle"><%= interpolate(
         gettext('Start Date: %(date)s'),
         { date: start_date },
         true
-    ) %></p>
+    ) %></div>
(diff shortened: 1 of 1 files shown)


schenedx:
@jimabramson Could you put a screenshot in the PR so I can see the UI change?

jimabramson:
@schenedx  hyg:
screen shot 2016-02-18 at 8 31 20 am
FYI @marcotuts

schenedx:
:+1:

programs/static/templates/course_run.underscore
```diff
@@ -21,6 +21,10 @@
         gettext('Start Date: %(date)s'),
         { date: start_date },
         true
+    ) %><br/><%= interpolate(
```
AlasdairSwan: I don't like the ```<br/>```, could we use CSS or a new ```<p>``` to create the same effect?

jimabramson:
changed to use two divs instead of a p with a br.   looks a little nicer IMO
screen shot 2016-02-19 at 10 50 23 am

AlasdairSwan:
Once tests pass :+1:

PR #103, opened 2016-02-23
Make access token optional when running sync_orgs command

rlucioni:
Takes advantage of the client credentials grant to get an access token when one isn't provided. ECOM-3349.

@bderusha please review.

### DIFF
--- a/programs/apps/programs/management/commands/sync_orgs.py
+++ b/programs/apps/programs/management/commands/sync_orgs.py
@@ -25,7 +25,10 @@ class Command(BaseCommand):
 
     def add_arguments(self, parser):
         parser.add_argument(
-            'access_token',
+            '-a', '--access_token',
+            action='store',
+            dest='access_token',
+            default=None,
             help='OAuth 2.0 access token used to authenticate against LMS APIs.'
         )
 
@@ -66,6 +69,23 @@ class Command(BaseCommand):
     def handle(self, *args, **options):
         access_token = options.get('access_token')
         commit = options.get('commit')
+
+        if access_token is None:
+            try:
+                access_token_url = '{}/access_token'.format(
+                    settings.SOCIAL_AUTH_EDX_OIDC_URL_ROOT.strip('/')
+                )
+                client_id = settings.SOCIAL_AUTH_EDX_OIDC_KEY
+                client_secret = settings.SOCIAL_AUTH_EDX_OIDC_SECRET
+
+                access_token, __ = EdxRestApiClient.get_oauth_access_token(
+                    access_token_url,
+                    client_id,
+                    client_secret
+                )
+            except:  # pylint: disable=bare-except
+                logger.exception('Unable to exchange client credentials grant for an access token.')
+                return
 
         self.client = EdxRestApiClient(settings.ORGANIZATIONS_API_URL_ROOT, oauth_access_token=access_token)
 
--- a/programs/apps/programs/tests/test_sync_orgs.py
+++ b/programs/apps/programs/tests/test_sync_orgs.py
@@ -1,10 +1,11 @@
 # pylint: disable=missing-docstring
+import itertools
 import json
 import math
 
 import ddt
 from django.conf import settings
-from django.core.management import call_command, CommandError
+from django.core.management import call_command
 from django.test import TestCase
 import httpretty
 
@@ -12,6 +13,7 @@ from programs.apps.programs.models import Organization
 
 
 @ddt.ddt
+@httpretty.activate
 class SyncOrgsTests(TestCase):
     """Tests for the sync_orgs management command."""
     ORGANIZATIONS_API_URL = settings.ORGANIZATIONS_API_URL_ROOT.strip('/') + '/organizations/'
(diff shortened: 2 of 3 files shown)


programs/apps/programs/tests/test_sync_orgs.py
```diff
@@ -62,21 +63,51 @@ def _mock_organizations_api(self):
             ]
         )
 
-    def test_handle_access_token_missing(self):
-        with self.assertRaises(CommandError):
-            call_command('sync_orgs')
+    def _mock_oauth2_provider(self, data=None):
```
bderusha: Since we are using a fully tested external dependency seems simpler to just mock the `get_oauth_access_token` method instead.

programs/apps/programs/tests/test_sync_orgs.py
```diff
@@ -62,21 +63,51 @@ def _mock_organizations_api(self):
...
+            content_type='application/json'
+        )
 
-    @ddt.data(True, False)
+    @ddt.data(*itertools.product(
+        ('fake-access-token', None),
+        (True, False)
+    ))
+    @ddt.unpack
     @httpretty.activate
```
bderusha: Can we call the `@httpretty.activate` decorator on the class since it applies to all methods?

programs/apps/programs/management/commands/sync_orgs.py
```diff
@@ -67,6 +70,23 @@ def handle(self, *args, **options):
...
 
+        if access_token is None:
+            try:
+                access_token_url = '{}/access_token'.format(
+                    settings.SOCIAL_AUTH_EDX_OIDC_URL_ROOT.strip('/')
+                )
+                client_id = settings.SOCIAL_AUTH_EDX_OIDC_KEY
+                client_secret = settings.SOCIAL_AUTH_EDX_OIDC_SECRET
+
+                access_token, _expires_at = EdxRestApiClient.get_oauth_access_token(
```
bderusha: Are we consistent about using `_unused_variable_name` as a pattern?  Or is it more clear to just use `__`

programs/apps/programs/management/commands/sync_orgs.py
```diff
@@ -67,6 +70,23 @@ def handle(self, *args, **options):
         access_token = options.get('access_token')
         commit = options.get('commit')
 
+        if access_token is None:
+            try:
+                access_token_url = '{}/access_token'.format(
+                    settings.SOCIAL_AUTH_EDX_OIDC_URL_ROOT.strip('/')
```
bderusha: :rocket: I have similar code in course-discovery but without the `strip('/')`.  I will add that in.

bderusha:
@rlucioni can you add a test case covering the raised Exception path for the client_credentials?

programs/apps/programs/tests/test_sync_orgs.py
rlucioni: I prefer using `httpretty` where possible because it allows more code to run than if I were to mock out `get_oauth_access_token()`. If that method is mocked and the package somehow changes underneath us, causing the method to behave differently, the tests won't be able to identify any issues.

programs/apps/programs/tests/test_sync_orgs.py
rlucioni: Can we? I didn't realize it worked that way.

rlucioni: It works! For reference, https://github.com/gabrielfalcao/HTTPretty/blob/master/httpretty/core.py#L1130.

programs/apps/programs/management/commands/sync_orgs.py
rlucioni: We're pretty consistent about it, as far as I can tell.

Typically, a single underscore is used to denote a throwaway variable. However, I don't like doing that. The single underscore is a special character in the Python shell used to store the last value. We also frequently import ugettext as `_`. In addition, the lone underscore may not mean anything to someone new to the code. I don't like forcing people to guess.

Selected passages from The Zen of Python, as a reference:
```python
>>> import this
The Zen of Python, by Tim Peters

Beautiful is better than ugly.
Explicit is better than implicit.
...
Readability counts.
```

programs/apps/programs/management/commands/sync_orgs.py
rlucioni: Good idea, feel free to tag me for review.

rlucioni:
@bderusha the path you're referring to is covered by `test_handle_access_token_unavailable()`.

programs/apps/programs/tests/test_sync_orgs.py
bderusha: For the record, I can get behind that and am fine leaving this as is.  But just to play devil's advocate:

We are using a versioned package so this does not enter into the realm of impacting us unless we decide to upgrade a major version bump.  One could argue that if we are worried about the contract of a 3rd party API changing we should test that contract explicitly rather than implicitly testing it through unit tests with different purposes.  To me this level of mocking muddies the test and makes it slightly harder to understand/maintain because it requires a detailed knowledge of what the 3rd party method is doing under the hood rather than just understanding what its inputs and outputs are.

programs/apps/programs/management/commands/sync_orgs.py
bderusha: "I don't like forcing people to guess."  That's exactly why I like using `__` (double underscore).  It doesn't conflict with single `_` ugettext, but it still clearly denotes that it is a throwaway variable.  Given the throwaway variable convention, IMO having an unused named variable to me forces people to guess more about why it's there.

bderusha:
:+1:

programs/apps/programs/tests/test_sync_orgs.py
rlucioni: I agree that this kind of mocking requires an understanding of the package beyond what would typically be required to use it. That said, I don't think it's necessarily a bad thing, especially since we're the ones maintaining the package.

Despite the fact that we're using a versioned package, what if the version gets bumped for use elsewhere in the application and the person making the change doesn't realize that the method is used here?

I'm on the other side of the fence as far as what this does for the cleanliness of the code. Isn't it cleaner to allow the code to run as it normally would, only interfering at the transport layer?

programs/apps/programs/management/commands/sync_orgs.py
rlucioni: I see what you're saying. Don't you think we lose some potentially valuable information when using `__`? Isn't the ideal knowing what the variable is meant to hold, but also that it's not used? What about `unused_expires_at`?

bderusha: To me that is what documentation is for.  Take the following example:

```python
def get_data():
    return {
        "name": "Renzo",
        "url": "http://instantcena.com"
        "org": "edX",
        "team": "ecom"
    }

data = get_data()
name = data['name']
url = data['url']
org = data['org']
team = data['team']

concat_org_and_team(org, team)
```

```python
def get_data():
    return "Renzo", "http://instantcena.com", "edX", "ecom"

name, url, org, team = get_data()

concat_org_and_team(org, team)
```

I don't think anyone would argue that there is no need to pull out `name` and `url` in the first example and if that's true then I don't see why it is necessary in the second either.

rlucioni: I think I see what you're getting after with your examples. In this case, the API client [returns a tuple](https://github.com/edx/edx-rest-api-client/blob/master/edx_rest_api_client/client.py#L44), so we're forced to unpack it here. We have to put the throwaway value somewhere. Would your ideal version of the second example look like this?
```python
def get_data():
    return "Renzo", "http://instantcena.com", "edX", "ecom"

__, __, org, team = get_data()

concat_org_and_team(org, team)
```

programs/apps/programs/tests/test_sync_orgs.py
bderusha: Yes, but it isn't a unit test to me anymore.  Which is fine, and this system level of integration testing is useful for exactly the reason you described.

But what if the OAuth spec changes and now we need to make a `PUT` request to get the token.  This test will fail because we are mocking the guts of an external package, the whole point of which is to abstract away that level of granular change.

The true answer is we need both, and lacking both it is safer to use your approach because there are fewer false-positives but the test is more brittle.

programs/apps/programs/management/commands/sync_orgs.py
bderusha: Yes, exactly.

programs/apps/programs/tests/test_sync_orgs.py
rlucioni: Understood. I think the OAuth spec is less likely to change than our package, so I'll stick with this, albeit conscious of the tradeoff.

programs/apps/programs/management/commands/sync_orgs.py
rlucioni: Updated.

PR #105, opened 2016-02-25
ECOM-3123 Acceptance test for Programs authoring app in Studio

schenedx:
@jimabramson @rlucioni I want you guys to start review this work while I work on documentation and incorporate these into build.

I believe this PR is ready for full review and merge for it's current state.
As for leverage the acceptance tests for build process, that's now a separate story at ECOM-3822

### DIFF
--- a/acceptance_tests/config.py
+++ b/acceptance_tests/config.py
@@ -1,0 +1,40 @@
+# pylint: disable=missing-docstring
+import os
+
+
+def str2bool(s):
+    s = unicode(s)
+    return s.lower() in (u'yes', u'true', u't', u'1')
+
+# TEST CONFIGURATION
+ALLOW_DELETE_ALL_PROGRAM = str2bool(os.environ.get('ALLOW_DELETE_ALL_PROGRAM', False))
+PROGRAM_ORGANIZATION = os.environ.get('PROGRAM_ORGANIZATION', 'edx')
+# END TEST CONFIGURATION
+
+# PROGRAM CONFIGURATION
+try:
+    PROGRAMS_URL_ROOT = os.environ.get('PROGRAMS_URL_ROOT').strip('/')
+except AttributeError:
+    raise RuntimeError('A valid url root for the Programs service must be provided to run acceptance tests.')
+# END PROGRAM CONFIGURATION
+
+# STUDIO CONFIGURATION
+try:
+    STUDIO_URL_ROOT = os.environ.get('STUDIO_URL_ROOT').strip('/')
+except AttributeError:
+    raise RuntimeError('You must provide a valid URL root for Studio to run acceptance tests.')
+
+STUDIO_EMAIL = os.environ.get('STUDIO_EMAIL')
+STUDIO_PASSWORD = os.environ.get('STUDIO_PASSWORD')
+BASIC_AUTH_USERNAME = os.environ.get('BASIC_AUTH_USERNAME')
+BASIC_AUTH_PASSWORD = os.environ.get('BASIC_AUTH_PASSWORD')
+# END STUDIO CONFIGURATION
+
+# LMS CONFIGURATION
+try:
+    LMS_URL_ROOT = os.environ.get('LMS_URL_ROOT').strip('/')
+except AttributeError:
+    raise RuntimeError('A valid url root for the LMS must be provided to run acceptance tests')
+
+LMS_AUTO_AUTH = str2bool(os.environ.get('LMS_AUTO_AUTH', True))
+# END LMS CONFIGURATION
--- a/acceptance_tests/mixins.py
+++ b/acceptance_tests/mixins.py
@@ -1,0 +1,147 @@
+# pylint: disable=missing-docstring
+import logging
+import requests
+import uuid
+
+from edx_rest_api_client.client import EdxRestApiClient
+
+from acceptance_tests.config import (
+    STUDIO_PASSWORD,
+    STUDIO_EMAIL,
+    BASIC_AUTH_USERNAME,
+    BASIC_AUTH_PASSWORD,
+    LMS_URL_ROOT,
+    PROGRAMS_URL_ROOT,
+    LMS_AUTO_AUTH)
+from acceptance_tests.pages import (
+    StudioLoginPage,
+    StudioHomePage,
+    StudioNewProgramPage,
+    StudioManageProgramPage,
+    StudioProgramIdTokenPage)
+
+log = logging.getLogger(__name__)
+
+
+class StudioUserMixin(object):
+    password = 'edx'
+
+    def generate_user_credentials(self, username_prefix):
+        username = username_prefix + uuid.uuid4().hex[0:10]
+        password = self.password
+        email = '{}@example.com'.format(username)
+        return username, email, password
+
+    def create_studio_user(self, is_staff=True):
+        username, email, password = self.generate_user_credentials(username_prefix='programs_acceptance')
+
+        url = ('{host}/auto_auth?no_login=true'
+               '&username={username}'
+               '&password={password}'
+               '&email={email}'
+               '&staff={staff}').format(host=LMS_URL_ROOT,
+                                        username=username,
+                                        password=password,
+                                        email=email,
+                                        staff='true' if is_staff else 'false')
+
+        auth = None
+
+        if BASIC_AUTH_USERNAME and BASIC_AUTH_PASSWORD:
+            auth = (BASIC_AUTH_USERNAME, BASIC_AUTH_PASSWORD)
+
+        requests.get(url, auth=auth)
+
+        return username, password, email
+
+
+class LogistrationMixin(StudioUserMixin):
+    def setUp(self):
+        super(LogistrationMixin, self).setUp()
+        self.studio_login_page = StudioLoginPage(self.browser)
+
+    def login(self, is_staff=True):
+        if STUDIO_EMAIL and STUDIO_PASSWORD:
+            # If the credential is specified through config, always use the specified credentials
+            self.login_with_studio(STUDIO_EMAIL, STUDIO_PASSWORD)
+        elif LMS_AUTO_AUTH:
+            # Only use auto auth if the config allows it
+            _, password, email = self.create_studio_user(is_staff)
+            self.login_with_studio(email, password)
+        else:
+            raise RuntimeError(
+                'User credentials have not been provided, and auto auth is not enabled.')
+
+    def login_with_studio(self, email, password):
+        # Note: We use Selenium directly here (as opposed to bok-choy) to avoid issues with promises being broken.
+        self.studio_login_page.browser.get(self.studio_login_page.url)
+        self.studio_login_page.login(email, password)
+
+
+class ProgramManagementMixin(StudioUserMixin):
+    def setUp(self):
+        super(ProgramManagementMixin, self).setUp()
+        self.session = requests.Session()
+        self.studio_home_page = StudioHomePage(self.browser)
+        self.manage_program_page = StudioManageProgramPage(self.browser)
+        self.program_token_page = StudioProgramIdTokenPage(self.browser)
+
+    def _get_program_jwt_token(self):
+        # Get the JWT token from the studio programs id_token page
+        self.program_token_page.visit()
+        token = self.program_token_page.get_token()
+        # Once we get the JWT token, we should go back to the page we
+        # came from to maintain the existing state of the test
+        self.browser.back()
+        return token
+
+    def _get_program_id_from_url(self, url):
+        program_id = url.split('/')[-1]
+        return program_id
+
+    def _get_current_program_id(self):
+        self.assertTrue(self.manage_program_page.is_browser_on_page())
+        # Get the newly created program id from the current url
+        return self._get_program_id_from_url(self.browser.current_url)
+
+    def create_program(self, name, description, marketing_slug):
+        self.studio_home_page.click_new_program()
+        new_program_page = StudioNewProgramPage(self.browser)
+        new_program_page.wait_for_page()
+
+        # Add the unique program info suffix
+        # to make sure we do not have collisions in db
+        program_unique_suffix = uuid.uuid4()
+
+        new_program_page.create_new_program(
+            '{}_{}'.format(name, program_unique_suffix)[:255],
+            '{} {}'.format(description, program_unique_suffix)[:255],
+            '{}_{}'.format(marketing_slug, program_unique_suffix)[:255])
+
+        self.manage_program_page.wait_for_page()
+
+    def _delete_program(self, program_id, jwt_token):
+        """ With the JWT token, hit the program details URL with the patch to set the
+            program status to "deleted". This is the delete program step """
+
+        url = '{0}/api/v1/'.format(PROGRAMS_URL_ROOT)
+        delete_client = EdxRestApiClient(url, jwt=jwt_token)
+        deleted_program = delete_client.programs(program_id).patch({'status': 'deleted'})
+        # tell the caller wither the delete is successful or not.
+        return deleted_program['status'] == 'deleted'
+
+    def delete_all_programs(self):
+        jwt_token = self._get_program_jwt_token()
+        # Make sure we are on the view programs page
+        self.assertTrue(self.studio_home_page.is_browser_on_page())
+        # Get all the program links from the page, and delete programs one by one
+        for program_link in self.studio_home_page.get_all_program_links():
+            program_id = self._get_program_id_from_url(program_link.get_attribute('href'))
+            if not self._delete_program(program_id, jwt_token):
+                log.warning('Set program %s to delete failed using jwt_token %s', program_id, jwt_token)
+        return True
+
+    def delete_current_program(self):
+        program_id = self._get_current_program_id()
+        jwt_token = self._get_program_jwt_token()
+        self.assertTrue(self._delete_program(program_id, jwt_token))
--- a/acceptance_tests/pages.py
+++ b/acceptance_tests/pages.py
@@ -1,0 +1,168 @@
+# pylint: disable=missing-docstring
+import abc
+import json
+
+from bok_choy.page_object import PageObject
+from bok_choy.promise import EmptyPromise
+from selenium.webdriver.support.select import Select
+
+from acceptance_tests.config import (
+    BASIC_AUTH_USERNAME,
+    BASIC_AUTH_PASSWORD,
+    STUDIO_URL_ROOT,
+    PROGRAM_ORGANIZATION,
+)
+
+
+class StudioPage(PageObject):  # pylint: disable=abstract-method
+    __metaclass__ = abc.ABCMeta
+
+    def _build_url(self, path):
+        url = '{}/{}'.format(STUDIO_URL_ROOT, path)
+
+        if BASIC_AUTH_USERNAME and BASIC_AUTH_PASSWORD:
+            url = url.replace('://', '://{}:{}@'.format(BASIC_AUTH_USERNAME, BASIC_AUTH_PASSWORD))
+
+        return url
+
+    def _hide_debug_tool_bar(self):
+        if self.q(css='a#djHideToolBarButton').visible:
+            # Hide the debug tool bar
+            self.q(css='a#djHideToolBarButton').click()
+
+    def _is_browser_on_studio_home(self):
+        return self.browser.title.startswith('Studio Home')
+
+
+class StudioRootPage(StudioPage):
+    @property
+    def url(self):
+        return self._build_url('')
+
+    def is_browser_on_page(self):
+        return self.browser.title.startswith('Welcome | Studio')
+
+
+class StudioLoginPage(StudioPage):
+    @property
+    def url(self):
+        return self._build_url('signin')
+
+    def is_browser_on_page(self):
+        return self.q(css='form#login_form').visible
+
+    def login(self, email, password):
+        self.q(css='input#email').fill(email)
+        self.q(css='input#password').fill(password)
+        self.q(css='button.action-primary').click()
+
+        # Wait for LMS to redirect to the dashboard
+        EmptyPromise(self._is_browser_on_studio_home, "Studio login redirected to dashboard").fulfill()
+
+
+class StudioHomePage(StudioPage):
+    @property
+    def url(self):
+        return self._build_url('home')
+
+    def is_browser_on_page(self):
+        return self._is_browser_on_studio_home()
+
+    def view_programs(self):
+        self.wait_for_element_visibility('ul#course-index-tabs', 'course index tab is now visible')
+        # make sure we have access to the programs tab
+        if self.q(css='li.programs-tab a').visible:
+            # Click on the programs tab and
+            self.q(css='li.programs-tab a').click()
+            # make sure the programs tab is now visible
+            self.wait_for_element_visibility('div.programs-tab', 'programs-tab is visible')
+            return True
+        else:
+            self.wait_for_element_absence(
+                'header.has-actions a.new-program-button',
+                'new program button is not present')
+            return False
+
+    def click_new_program(self):
+        # This function would click on the "New Program" button on the header section
+        # to go to the new program page on studio.
+        self._hide_debug_tool_bar()
+        self.wait_for_element_presence('header.has-actions a.new-program-button', 'new program button is present')
+        self.q(css='header.has-actions a.new-program-button').click()
+
+    def get_all_program_links(self):
+        self.view_programs()
+        return self.q(css='a.program-link')
+
+    def is_program_list_empty(self):
+        self.view_programs()
+        self.wait_for_element_visibility(
+            'div.programs-tab a.action-create-program',
+            'create new program button in program tab visible')
+        return True
+
+
+class StudioNewProgramPage(StudioPage):
+    @property
+    def url(self):
+        return self._build_url('program/new')
+
+    def is_browser_on_page(self):
+        return self.browser.title.startswith('Program Administration') and \
+            self.q(css='button.js-create-program').visible
+
+    def create_new_program(self, program_name, program_subtitle, program_marketing_slug):
+        self.q(css='input#program-name').fill(program_name)
+        self.q(css='input#program-subtitle').fill(program_subtitle)
+        self.q(css='input#program-marketing-slug').fill(program_marketing_slug)
+        select = Select(self.browser.find_element_by_css_selector('select#program-org'))
+        select.select_by_value(PROGRAM_ORGANIZATION)
+        # Click on create program button to create the new program
+        self.q(css='button.js-create-program').first.click()
+
+
+class StudioManageProgramPage(StudioPage):
+    # Not going to navigate to this page directly
+    url = None
+
+    def is_browser_on_page(self):
+        return self.q(css='div.js-inline-edit').visible
+
+    def add_course(self, course_code, course_title):
+        self.wait_for_element_visibility('button.js-add-course', 'add course button visible')
+        self.q(css='button.js-add-course').first.click()
+        self.wait_for_element_visibility('input.course-key', 'course code field is visible')
+        self.q(css='input.course-key').first.fill(course_code)
+        self.q(css='input.display-name').first.fill(course_title)
+        self.q(css='button.js-select-course').first.click()
+
+    def add_course_run(self):
+        self.wait_for_element_visibility('button.js-add-course-run', 'add course run button visible')
+        self.q(css='button.js-add-course-run').first.click()
+        self.wait_for_element_visibility(
+            'select.js-course-run-select option[value*=\'course\']', 'select course options visible')
+        select = Select(self.browser.find_element_by_css_selector('select.js-course-run-select'))
+        select.select_by_index(1)
+
+        self.wait_for_element_visibility('button.js-remove-run', 'remove course run button visible')
+
+    def publish_program(self):
+        self.wait_for_element_visibility('button.js-publish-program', 'program public button visible')
+        self.q(css='button.js-publish-program').first.click()
+        self.wait_for_element_visibility('div.confirm-modal button.js-confirm', 'confirmation modal visible')
+        self.q(css='div.confirm-modal button.js-confirm').first.click()
+        self.wait_for_element_absence('button.js-publish-program', 'publish program button invisible')
+
+
+class StudioProgramIdTokenPage(StudioPage):
+    @property
+    def url(self):
+        return self._build_url('programs/id_token')
+
+    def is_browser_on_page(self):
+        return self.q(css='pre').visible
+
+    def get_token(self):
+        token_string = self.q(css='pre').first.text[0]
+        token = json.loads(token_string)
+        return token['id_token']
--- a/acceptance_tests/test_programs.py
+++ b/acceptance_tests/test_programs.py
@@ -1,0 +1,52 @@
+# pylint: disable=missing-docstring
+from unittest import skipUnless
+
+from bok_choy.web_app_test import WebAppTest
+
+from acceptance_tests.mixins import LogistrationMixin, ProgramManagementMixin
+from acceptance_tests.config import ALLOW_DELETE_ALL_PROGRAM, LMS_AUTO_AUTH
+
+
+class ProgramTests(LogistrationMixin, ProgramManagementMixin, WebAppTest):
+
+    def test_view_programs(self):
+        self.login()
+        self.assertTrue(self.studio_home_page.view_programs())
+
+    def test_create_program(self):
+        self.login()
+        self.create_program(
+            'testProgram',
+            'This is a test program',
+            'test_program_marketing_slug')
+
+        # Do program data cleanup
+        self.delete_current_program()
+
+    def test_publish_program(self):
+        self.login()
+        self.create_program(
+            'testProgramWithCourse',
+            'This is a test program with course',
+            'Test_Course_Program')
+        try:
+            self.manage_program_page.add_course(
+                'EDX Demo Course',
+                'edx demo')
+
+            self.manage_program_page.add_course_run()
+            self.manage_program_page.publish_program()
+        finally:
+            # We should cleanup the program data regardless if the test suceeded or failed.
+            self.delete_current_program()
+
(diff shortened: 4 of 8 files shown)


programs/apps/api/v1/tests/test_views.py
```diff
@@ -237,6 +237,9 @@ def test_create_with_invalid_org_data(self):
         {'subtitle': 'dummy-subtitle'},
         {'marketing_slug': 'dummy-marketing-slug'},
         {'subtitle': 'dummy-subtitle', 'marketing_slug': 'dummy-marketing-slug'},
+        {'status': 'unpublished'},
```
schenedx: I added these extra unit test cases to make sure Program status updates through API works

acceptance_tests/pages.py
```diff
@@ -0,0 +1,172 @@
...
+        self.wait_for_element_visibility('ul#course-index-tabs', 'course index tab is now visible')
+        # make sure we have access to the programs tab
+        if self.q(css='li.programs-tab a').visible:
+            # Click on the programs tab and
+            self.q(css='li.programs-tab a').click()
+            # make sure the programs tab is now visible
+            self.wait_for_element_visibility('div.programs-tab', 'programs-tab is visible')
+            return True
+        else:
+            self.wait_for_element_absence('header.has-actions a.new-program-button', 'new program button is not present')
```
jimabramson: Is it deliberate that you do not check for the presence of this element when the tab is visible?

acceptance_tests/test_programs.py
```diff
@@ -0,0 +1,58 @@
+from bok_choy.web_app_test import WebAppTest
+from acceptance_tests.mixins import LogistrationMixin, ProgramManagementMixin
+from acceptance_tests.config import ALLOW_DELETE_ALL_PROGRAM
+
+
+class ProgramTests(LogistrationMixin, ProgramManagementMixin, WebAppTest):
+    def setUp(self):
+        super(ProgramTests, self).setUp()
```
jimabramson: not sure if you have plans to add to this setUp method, but as of now, it doesn't seem necessary.

acceptance_tests/pages.py
schenedx: Yes. This function is also used by the non staff user. For non staff user, the programs-tab is not visible. And here, I am also explicitly checking if the create program button is not present. If these are true, I am returning false to say, "cannot view program"

jimabramson: OK

acceptance_tests/pages.py
```diff
@@ -0,0 +1,172 @@
...
+            return False
+
+    def new_program(self):
+        # This function would click on the "New Program" button on the header section
+        # to go to the new program page on studio.
+        if self.q(css='a#djHideToolBarButton').visible:
+            #Hide the debug tool bar
+            self.q(css='a#djHideToolBarButton').click()
```
jimabramson: I'd suggest moving this part into a helper method since you might want to ensure the tollbar is disabled within the context of any page operation.

acceptance_tests/pages.py
```diff
@@ -0,0 +1,172 @@
...
+            # Click on the programs tab and
+            self.q(css='li.programs-tab a').click()
+            # make sure the programs tab is now visible
+            self.wait_for_element_visibility('div.programs-tab', 'programs-tab is visible')
+            return True
+        else:
+            self.wait_for_element_absence('header.has-actions a.new-program-button', 'new program button is not present')
+            return False
+
+    def new_program(self):
```
jimabramson: it would more clearly signal intent to have a verb in the name here, like `click_new_program`

acceptance_tests/mixins.py
```diff
@@ -0,0 +1,174 @@
...
+        # Make sure we are on the view programs page
+        self.assertTrue(self.studio_home_page.is_browser_on_page())
+        # Get all the program links from the page, and delete programs one by one
+        for program_link in self.studio_home_page.get_all_program_links():
+            program_id = self._get_program_id_from_url(program_link.get_attribute('href'))
+            if not self._delete_program(program_id, jwt_token):
+                raise RuntimeError('Delete program with id {} failed'.format(program_id))
+        return True
+
+    def program_data_cleanup(self):
```
jimabramson: suggest renaming to `delete_current_program` or similar, to better disambiguate from `delete_all_programs`

acceptance_tests/test_programs.py
```diff
@@ -0,0 +1,58 @@
...
+            self.manage_program_page.add_course(
+                'EDX Demo Course',
+                'edx demo')
+        
+            self.manage_program_page.add_course_run(
+                'course-v1:edX+DemoX+Demo_Course')
+            self.manage_program_page.publish_program()
+
+        except Exception as e:
+            raise RuntimeError('publish program encountered errors: {}'.format(e))
```
jimabramson: the underlying traceback will be lost here.  maybe it would be better to let the original exception go uncaught?

acceptance_tests/pages.py
```diff
@@ -0,0 +1,172 @@
...
+            self.q(css='a#djHideToolBarButton').click()
+        self.wait_for_element_presence('header.has-actions a.new-program-button', 'new program button is present')
+        self.q(css='header.has-actions a.new-program-button').click()
+
+    def get_all_program_links(self):
+        self.view_programs()
+        return self.q(css='a.program-link')
+
+
+    def no_program(self):
```
jimabramson: another rename suggestion for clarity, maybe `view_empty_program_list` ?

acceptance_tests/test_programs.py
schenedx: Agree, will remove

acceptance_tests/pages.py
schenedx: Good point, will do.

acceptance_tests/pages.py
schenedx: Will update the function name to `click_new_program`

acceptance_tests/mixins.py
schenedx: Agree with the function name change.

acceptance_tests/test_programs.py
schenedx: The reason I want to do this is because I wan to make sure the code in the `final` block is executed. Do you see if we remove the `except` here, we can achieve that still?

acceptance_tests/pages.py
schenedx: Agree and will change

acceptance_tests/test_programs.py
rlucioni: Code in a `finally` block is always executed "on the way out," even when an exception is raised.

rlucioni: For reference, see https://docs.python.org/2/tutorial/errors.html#defining-clean-up-actions.

schenedx: Got it. Updated.

programs/apps/api/v1/tests/test_views.py
jimabramson: good call.

acceptance_tests/test_programs.py
```diff
@@ -0,0 +1,52 @@
...
+            self.manage_program_page.publish_program()
+        finally:
+            # We should cleanup the program data regardless if the test suceeded or failed.
+            self.delete_current_program()
+
+    def test_non_staff_user(self):
+        self.login(is_staff=False)
+        self.assertFalse(self.studio_home_page.view_programs())
+
+    def test_view_no_program(self):
```
jimabramson: not a big deal, but if WebAppTest is based on unittest.TestCase, then [skipUnless](https://docs.python.org/2/library/unittest.html#unittest.skipUnless) might be a good option here.

acceptance_tests/pages.py
```diff
@@ -0,0 +1,174 @@
...
+
+        return url
+
+    def _hide_debug_tool_bar(self):
+        if self.q(css='a#djHideToolBarButton').visible:
+            # Hide the debug tool bar
+            self.q(css='a#djHideToolBarButton').click()
+
+    def _is_browser_on_studio_home(self):
+        return lambda: self.browser.title.startswith('Studio Home')
```
jimabramson: simplification: could get rid of the lambda part here, and just pass the method (without calling it) down on L#58

jimabramson: actually this logic is redefined in `StudioHomePage` (but not as a lambda) - can these be consolidated?

acceptance_tests/test_programs.py
schenedx: Agree. looks much more readable too.

acceptance_tests/pages.py
schenedx: Agree. Will update

requirements/test.txt
```diff
@@ -15,3 +15,5 @@ nose-ignore-docstring==0.2
 pep8==1.6.2
 pylint==1.4.4
 testfixtures==4.4.0
+bok-choy==0.4.7
```
jzoldak: The latest is 0.4.10.
Unless there is a reason that it won't work, you should use that version. (And let me know so we can make any changes you need.)

schenedx: I just didn't know 0.4.10 is the latest version. Will update

acceptance_tests/config.py
```diff
@@ -0,0 +1,40 @@
...
+    PROGRAMS_URL_ROOT = os.environ.get('PROGRAMS_URL_ROOT').strip('/')
+except AttributeError:
+    raise RuntimeError('A valid url root for the Programs service must be provided to run acceptance tests.')
+# END PROGRAM CONFIGURATION
+
+# STUDIO CONFIGURATION
+try:
+    STUDIO_URL_ROOT = os.environ.get('STUDIO_URL_ROOT').strip('/')
+except AttributeError:
+    raise RuntimeError('You must provide a valid URL root for the Studio to run acceptance tests.')
```
rlucioni: s/the Studio/Studio

acceptance_tests/mixins.py
```diff
@@ -0,0 +1,168 @@
+# pylint: disable=missing-docstring
+import logging
+import uuid
+import json
+import re
```
rlucioni: Nitpick: Please alphabetize these. Pro-tip: in Sublime, you can highlight the lines and hit F5.

acceptance_tests/mixins.py
```diff
@@ -0,0 +1,168 @@
...
+
+import requests
+from urlparse import urlparse
```
rlucioni: Nitpick: urlparse is part of the standard library, so it should be included in the first block of imports.

acceptance_tests/mixins.py
```diff
@@ -0,0 +1,168 @@
...
+    def generate_user_credentials(self, username_prefix):
+        username = username_prefix + uuid.uuid4().hex[0:10]
+        password = self.password
+        email = '{}@example.com'.format(username)
+        return username, email, password
+
+    def create_studio_user(self, is_staff=True):
+        username, email, password = self.generate_user_credentials(username_prefix='programs_acceptance')
+
+        url = '{host}/auto_auth?no_login=true&username={username}&password={password}&email={email}&staff={staff}'.format(  # pylint: disable=C0301
```
rlucioni: Instead of ignoring Pylint, try wrapping this line. You can use parentheses to do this as you would for a long function call. Adjacent string literals are concatenated at compile time. For example:
```python
url = (
    '{host}/auto_auth?no_login=true'
    '&username={username}'
    '&password={password}'
    '&email={email}'
    '&staff={staff}'
).format(...)
```

acceptance_tests/mixins.py
```diff
@@ -0,0 +1,168 @@
...
+        if STUDIO_EMAIL and STUDIO_PASSWORD:
+            # If the credential is specified through config, always use the specified credentials
+            self.login_with_studio(STUDIO_EMAIL, STUDIO_PASSWORD)
+        elif LMS_AUTO_AUTH:
+            # Only use auto auth if the config allows it
+            _, password, email = self.create_studio_user(is_staff)
+            self.login_with_studio(email, password)
+        else:
+            raise RuntimeError(
+                'The test has no specified user credential and do not allow Auto Auth user creation')
```
rlucioni: "User credentials have not been provided, and auto auth is not enabled."

acceptance_tests/mixins.py
```diff
@@ -0,0 +1,168 @@
...
+
+    def login_with_studio(self, email=None, password=None):
+        """ Visit LMS and login. """
```
rlucioni: This docstring doesn't seem accurate. You're visiting Studio. I don't think this docstring is even necessary. The function name makes it clear (which is great).

acceptance_tests/mixins.py
```diff
@@ -0,0 +1,168 @@
...
+            self.login_with_studio(STUDIO_EMAIL, STUDIO_PASSWORD)
+        elif LMS_AUTO_AUTH:
+            # Only use auto auth if the config allows it
+            _, password, email = self.create_studio_user(is_staff)
+            self.login_with_studio(email, password)
+        else:
+            raise RuntimeError(
+                'The test has no specified user credential and do not allow Auto Auth user creation')
+
+    def login_with_studio(self, email=None, password=None):
```
rlucioni: Since `email` and `password` aren't optional, they should be positional arguments, not keyword arguments.

acceptance_tests/mixins.py
```diff
@@ -0,0 +1,168 @@
...
+        """ Visit LMS and login. """
+        # email must be specified
+        self.assertTrue(email is not None)
+        # password must be specified
+        self.assertTrue(password is not None)
```
rlucioni: I don't think these assertions are necessary.

acceptance_tests/mixins.py
```diff
@@ -0,0 +1,168 @@
...
+        self.session = requests.Session()
+        self.studio_home_page = StudioHomePage(self.browser)
+        self.manage_program_page = StudioManageProgramPage(self.browser)
+        self.program_token_page = StudioProgramIdTokenPage(self.browser)
+
+    def _get_program_jwt_token(self):
+        # Get the JWT token from the studio programs id_token page
+        self.program_token_page.visit()
+        token = self.program_token_page.get_token()
+        self.browser.back()
```
rlucioni: Could you add a comment about why this is necessary?

acceptance_tests/mixins.py
```diff
@@ -0,0 +1,168 @@
...
+        return token
+
+    def _get_program_id_from_url(self, url):
+        parsed_url = urlparse(url)
+        program_id = int(re.findall(r'\d+', parsed_url.path)[0])
```
rlucioni: More simply:
```python
program_id = url.split('/')[-1]
```

acceptance_tests/mixins.py
```diff
@@ -0,0 +1,168 @@
...
+        return self._get_program_id_from_url(self.browser.current_url)
+
+    def create_program(self, name, description, marketing_slug):
+        self.studio_home_page.click_new_program()
+        new_program_page = StudioNewProgramPage(self.browser)
+        new_program_page.wait_for_page()
+
+        # Add the unique program info suffix
+        # to make sure we do not have collisions in db
+        program_unique_suffix = str(uuid.uuid4())
```
rlucioni: You don't need to do this string conversion. It happens automatically when you interpolate a UUID into a string.

acceptance_tests/mixins.py
```diff
@@ -0,0 +1,168 @@
...
+
+        new_program_page.create_new_program(
+            '{}_{}'.format(name, program_unique_suffix),
+            '{} {}'.format(description, program_unique_suffix),
+            '{}_{}'.format(marketing_slug, program_unique_suffix))
```
rlucioni: I like this unique suffix idea, but I'd watch out for issues with these fields becoming too long as a result. Would it make sense to do some validation to make sure things that are too long get truncated?

acceptance_tests/mixins.py
```diff
@@ -0,0 +1,168 @@
...
+        # With the JWT token, hit the program details URL with the patch to set the
+        # program status to "deleted". This is the delete program step
+        url = '{0}/api/v1/programs/{1}/'.format(PROGRAMS_URL_ROOT, program_id)
+
+        patch_data = {'status': 'deleted'}
+
+        patch_response = requests.patch(
+            url,
+            data=json.dumps(patch_data),
+            headers={'Authorization': auth, 'Content-Type': 'application/merge-patch+json'})
```
rlucioni: This is a perfect use case for the [edx-rest-api-client](https://github.com/edx/edx-rest-api-client). It's already installed by Programs. Here's [an example](https://github.com/edx/credentials/blob/master/acceptance_tests/mixins.py#L50) of how you can use it with a JWT. Let me know if you have questions.

acceptance_tests/mixins.py
```diff
@@ -0,0 +1,168 @@
...
+
+    def _delete_program(self, program_id, jwt_token):
+
+        auth = 'JWT {0}'.format(jwt_token)
+
+        # With the JWT token, hit the program details URL with the patch to set the
+        # program status to "deleted". This is the delete program step
+        url = '{0}/api/v1/programs/{1}/'.format(PROGRAMS_URL_ROOT, program_id)
+
+        patch_data = {'status': 'deleted'}
```
rlucioni: What are the benefits of a soft-delete vs. a hard-delete? If we stick with the soft-delete, we're going to end up with a bunch of stale test programs in databases these tests are run against.

acceptance_tests/mixins.py
```diff
@@ -0,0 +1,168 @@
...
+
+    def delete_all_programs(self):
+        jwt_token = self._get_program_jwt_token()
+        # Make sure we are on the view programs page
+        self.assertTrue(self.studio_home_page.is_browser_on_page())
+        # Get all the program links from the page, and delete programs one by one
+        for program_link in self.studio_home_page.get_all_program_links():
+            program_id = self._get_program_id_from_url(program_link.get_attribute('href'))
+            if not self._delete_program(program_id, jwt_token):
+                raise RuntimeError('Delete program with id {} failed'.format(program_id))
```
rlucioni: Raising an exception doesn't seem quite right here. What if you're in the middle of deleting a long list of programs, and one fails? This will give up on all the others without trying to delete them first.

acceptance_tests/pages.py
```diff
@@ -0,0 +1,176 @@
+# pylint: disable=missing-docstring
+import abc
+import json
+from bok_choy.page_object import PageObject
```
rlucioni: Nitpick: please separate standard library imports from third-party imports. We should really update Pylint so I don't need to keep calling this out...

acceptance_tests/pages.py
```diff
@@ -0,0 +1,176 @@
...
+from acceptance_tests.config import (
+    BASIC_AUTH_USERNAME,
+    BASIC_AUTH_PASSWORD,
+    STUDIO_URL_ROOT,
+    PROGRAM_ORGANIZATION,
+)
+
+
+class StudioPage(PageObject):  # pylint: disable=abstract-method
+    __metaclass__ = abc.ABCMeta
```
rlucioni: This class is used as a mixin (which is fine) but it doesn't define any abstract methods. Does it need to be an ABC?

rlucioni: Whoops, ignore this. PageObject itself is an ABC.

acceptance_tests/pages.py
```diff
@@ -0,0 +1,176 @@
...
+
+    def _build_url(self, path):
+        url = '{}/{}'.format(STUDIO_URL_ROOT, path)
+
+        if BASIC_AUTH_USERNAME and BASIC_AUTH_PASSWORD:
+            url = url.replace('://', '://{}:{}@'.format(BASIC_AUTH_USERNAME, BASIC_AUTH_PASSWORD))
+
+        return url
+
+    def _hide_debug_tool_bar(self):
```
rlucioni: The tests should be able to just ignore this, right? Is there a reason the toolbar must be dismissed?

acceptance_tests/pages.py
```diff
@@ -0,0 +1,176 @@
...
+        return self.q(css='form#login_form').visible
+
+    def login(self, email, password):
+        self.q(css='input#email').fill(email)
+        self.q(css='input#password').fill(password)
+        self.q(css='button.action-primary').click()
+
+        # Wait for LMS to redirect to the dashboard
+        # pylint: disable=unnecessary-lambda
+        EmptyPromise(lambda: self._is_browser_on_studio_home(), "Studio login redirected to dashboard").fulfill()
```
rlucioni: I've done this before without a lambda. What happens if you get rid of it?

rlucioni: I see, it's because `_is_browser_on_studio_home ` isn't a lambda. Can't you just pass `self._is_browser_on_studio_home`, without calling it?

acceptance_tests/pages.py
```diff
@@ -0,0 +1,176 @@
...
+        # to go to the new program page on studio.
+        self._hide_debug_tool_bar()
+        self.wait_for_element_presence('header.has-actions a.new-program-button', 'new program button is present')
+        self.q(css='header.has-actions a.new-program-button').click()
+
+    def get_all_program_links(self):
+        self.view_programs()
+        return self.q(css='a.program-link')
+
+    def view_empty_program_list(self):
```
rlucioni: Given what it's doing, a more readable name for this method might be `is_program_list_empty`.

acceptance_tests/pages.py
```diff
@@ -0,0 +1,176 @@
...
+
+    def create_new_program(self, program_name, program_subtitle, program_marketing_slug):
+        self.q(css='input#program-name').fill(program_name)
+        self.q(css='input#program-subtitle').fill(program_subtitle)
+        self.q(css='input#program-marketing-slug').fill(program_marketing_slug)
+
+        try:
+            select = Select(self.browser.find_element_by_css_selector('select#program-org'))
+            select.select_by_value(PROGRAM_ORGANIZATION)
+        except:
```
rlucioni: Selenium raises a perfectly good exception when it can't find an option you've asked it to select ([source](http://selenium.googlecode.com/svn/trunk/docs/api/py/_modules/selenium/webdriver/support/select.html#Select.select_by_value)). I'd let that exception be raised instead of catching it and replacing it with something else.

acceptance_tests/pages.py
```diff
@@ -0,0 +1,176 @@
...
+
+        self.q(css='button.js-create-program').first.click()
+
+
+class StudioManageProgramPage(StudioPage):
+    # Not going to navigate to this page directly
+    url = None
+
+    def is_browser_on_page(self):
+        return self.q(css='div.js-inline-edit').visible
```
rlucioni: I'd look for `.js-program-admin` instead: https://github.com/edx/edx-platform/blob/master/cms/templates/program_authoring.html#L17

acceptance_tests/pages.py
```diff
@@ -0,0 +1,176 @@
...
+
+    def add_course_run(self):
+        self.wait_for_element_visibility('button.js-add-course-run', 'add course run button visible')
+        self.q(css='button.js-add-course-run').first.click()
+        self.wait_for_element_visibility(
+            'select.js-course-run-select option[value*=\'course\']', 'select course options visible')
+        try:
+            select = Select(self.browser.find_element_by_css_selector('select.js-course-run-select'))
+            select.select_by_index(1)
+        except Exception as e:
```
rlucioni: Again, I'd avoid doing this and instead let Selenium raise its own exception.

acceptance_tests/pages.py
```diff
@@ -0,0 +1,176 @@
...
+            raise RuntimeError(
+                'Not able to set course run and the exception is {}'.format(e))
+
+        self.wait_for_element_visibility('button.js-remove-run', 'remove course run button visible')
+
+    def publish_program(self):
+        self.wait_for_element_visibility('button.js-publish-program', 'program public button visible')
+        self.q(css='button.js-publish-program').first.click()
+        self.wait_for_element_visibility('div.confirm-modal button.js-confirm', 'confirmation model visible')
```
rlucioni: s/model/modal

acceptance_tests/test_programs.py
```diff
@@ -0,0 +1,51 @@
+# pylint: disable=missing-docstring
+from bok_choy.web_app_test import WebAppTest
+from acceptance_tests.mixins import LogistrationMixin, ProgramManagementMixin
+from acceptance_tests.config import ALLOW_DELETE_ALL_PROGRAM, LMS_AUTO_AUTH
+from unittest import skipUnless
```
rlucioni: Nitpick: Please separate your imports into the standard three blocks.

acceptance_tests/mixins.py
```diff
@@ -0,0 +1,168 @@
...
+        # email must be specified
+        self.assertTrue(email is not None)
+        # password must be specified
+        self.assertTrue(password is not None)
+        # Note: We use Selenium directly here (as opposed to bok-choy) to avoid issues with promises being broken.
+        self.studio_login_page.browser.get(self.studio_login_page.url())  # pylint: disable=not-callable
+        self.studio_login_page.login(email, password)
+
+
+class LogoutMixin(object):
```
rlucioni: This is unused.

requirements/test.txt
```diff
@@ -15,3 +15,5 @@ nose-ignore-docstring==0.2
 pep8==1.6.2
 pylint==1.4.4
 testfixtures==4.4.0
+bok-choy==0.4.10
+selenium>=2.48.0
```
rlucioni: Any reason not to use the latest 2.52.0? Also, please pin (i.e., use `==`) to the latest working version. Otherwise, this requirement will change underneath us.

acceptance_tests/test_programs.py
```diff
@@ -0,0 +1,51 @@
...
+        self.delete_current_program()
+
+    def test_publish_program(self):
+        self.login()
+        self.create_program(
+            'testProgramWithCourse',
+            'This is a test program with course',
+            'Test_Course_Program')
+        try:
+
```
rlucioni: Nitpick: unnecessary line.

docs/testing.rst
```diff
@@ -12,3 +12,61 @@ Code quality validation can be run independently with:
 .. code-block:: bash
 
     $ make quality
+
+Acceptance Testing
+------------------
+
+The acceptance testing on this project mostly happens on the CMS system (Studio). The UI component of the system is presented
+as a seamless experience on the CMS. There are configuration of the environment these tests needs before the test can run
+which produces value on quality control.
```
rlucioni: Acceptance tests for this project are intended for use with Studio, where the Programs admin tool is used.

docs/testing.rst
```diff
@@ -12,3 +12,61 @@ Code quality validation can be run independently with:
...
+Our acceptance tests rely on configuration which are specified using environment variables below.
+
+======================== ========================================================================= ========= ============================================================
+Variable                 Description                                                               Required? Default Value
+======================== ========================================================================= ========= ============================================================
+STUDIO_URL_ROOT          URL root for the CMS platform Service                                     Yes       N/A
+LMS_URL_ROOT             URL root for the LMS platform Service                                     Yes       N/A
+PROGRAMS_URL_ROOT        URL root for the Programs Service                                         Yes       N/A
+STUDIO_EMAIL             The email of the user to log into CMS              					   maybe     N/A
+STUDIO_PASSWORD          The password of the user to log into CMS              					   maybe     N/A
```
rlucioni: Update the table to indicate that STUDIO_EMAIL and STUDIO_PASSWORD are not required. The have no default value. Your note below explains how these settings work when combined with LMS_AUTO_AUTH.

docs/testing.rst
```diff
@@ -12,3 +12,61 @@ Code quality validation can be run independently with:
...
+PROGRAM_ORGANIZATION     The organization to choose when creating the new program                  No        'edx'
+BASIC_AUTH_USERNAME      Username used to bypass HTTP basic auth on the LMS                        No        N/A
+BASIC_AUTH_PASSWORD      Password used to bypass HTTP basic auth on the LMS                        No        N/A
+======================== ========================================================================= ========= ============================================================
+
+Note:
+* If STUDIO_EMAIL and STUDIO_PASSWORD are used, the tests will always use the credentials for all the tests. 
+This also means the the acceptance tests will not use auto auth capability on LMS
+* If LMS_AUTO_AUTH is set to "False" and no STUDIO_EMAIL and STUDIO_PASSWORD are specified. 
+The tests will fail. This is because the tests needs user credentials to experience the Studio UI
```
rlucioni: This isn't rendering quite how you want it to. I think something like the following would be fine:
```
If STUDIO_EMAIL and STUDIO_PASSWORD are provided, they will be used for all tests. This means that the tests will not use auto-auth. If LMS_AUTO_AUTH is set to "False" and STUDIO_EMAIL and STUDIO_PASSWORD are not specified, the tests will fail. This is because the tests needs user credentials to access Studio.
```

rlucioni:
@schenedx I've completed my review. Let me know if you have questions.

acceptance_tests/mixins.py
jimabramson: @rlucioni the API doesn't support hard-deletion at present.  we can add that but i'd want to restrict its usage somehow.

acceptance_tests/mixins.py
jimabramson: agree that warning and continuing seems like a safer approach.

acceptance_tests/pages.py
jimabramson: i think it actually gets in the way of the new program button in studio

schenedx: Yes, that's the reason. It's super annoying and took me an hour to figure out. The debug toolbar would be on top of the "new programs+" button and selenium won't be able to click on it.

acceptance_tests/pages.py
schenedx: Good point. Will try

acceptance_tests/pages.py
schenedx: I believe `.js-program-admin` element shows also for StudioNewProgramPage, therefore that element is not a good indication whether the browser is on the program management page or not. I tried to use this element in my testing on this change and it did not work.

acceptance_tests/mixins.py
schenedx: I don't know it make sense to increase the scope of this ticket to implement `hard-delete` on Programs. I say we should try to use `soft-delete` in our tests for now. And if we found the data in the DB would be a problem later, we can always go back to it and use hard-delete. Let me create a jira ticket to capture this potential need.

schenedx: [ECOM-3837](https://openedx.atlassian.net/browse/ECOM-3837) created to track this need

acceptance_tests/pages.py
rlucioni: Huh, that's surprising. I didn't realize that could mess up Selenium. Thanks for the explanation.

acceptance_tests/mixins.py
rlucioni: Thanks @schenedx. For reference, this would be pretty easy to do with a DRF [DestroyModelMixin](http://www.django-rest-framework.org/api-guide/generic-views/#destroymodelmixin).

acceptance_tests/pages.py
rlucioni: @schenedx got it, thanks for the explanation.

schenedx:
@rlucioni @jimabramson I have address all your feedbacks. I have also squashed commits. Please review again.

acceptance_tests/mixins.py
```diff
@@ -0,0 +1,147 @@
...
+
+        new_program_page.create_new_program(
+            '{}_{}'.format(name, program_unique_suffix)[:255],
+            '{} {}'.format(description, program_unique_suffix)[:255],
+            '{}_{}'.format(marketing_slug, program_unique_suffix)[:255])
+
+        self.manage_program_page.wait_for_page()
+
+    def _delete_program(self, program_id, jwt_token):
+        '''With the JWT token, hit the program details URL with the patch to set the
```
rlucioni: Nitpick: as per [PEP-257](https://www.python.org/dev/peps/pep-0257/#what-is-a-docstring), please stick to triple-double quotes for docstrings.

docs/testing.rst
```diff
@@ -12,3 +12,56 @@ Code quality validation can be run independently with:
 .. code-block:: bash
 
     $ make quality
+
+Acceptance Testing
+------------------
+
+Acceptance tests for this project are intended for use with Studio, where the Programs admin tool is used. There are configuration environment configuration these tests need which we can maintain the quality of the service.
```
rlucioni: The second sentence here isn't necessary. Please remove it.

docs/testing.rst
```diff
@@ -12,3 +12,56 @@ Code quality validation can be run independently with:
...
+Our acceptance tests rely on configuration which are specified using environment variables below.
+
+======================== ========================================================================= ========= ============================================================
+Variable                 Description                                                               Required? Default Value
+======================== ========================================================================= ========= ============================================================
+STUDIO_URL_ROOT          URL root for the CMS platform Service                                     Yes       N/A
+LMS_URL_ROOT             URL root for the LMS platform Service                                     Yes       N/A
+PROGRAMS_URL_ROOT        URL root for the Programs Service                                         Yes       N/A
+STUDIO_EMAIL             The email of the user to log into CMS              					   NO        N/A
+STUDIO_PASSWORD          The password of the user to log into CMS              					   NO        N/A
```
rlucioni: Your additions to this table aren't rendering correctly. The spacing is important here. If you fix that, the table should render correctly.

acceptance_tests/pages.py
```diff
@@ -0,0 +1,176 @@
...
+    @property
+    def url(self):
+        return self._build_url('')
+
+    def is_browser_on_page(self):
+        return self.browser.title.startswith('Welcome | Studio')
+
+
+class StudioLoginPage(StudioPage):
+    def url(self):
```
rlucioni: Can this be a property, as you have below?
```python
@property
def url(self):
    return self._build_url('signin')
```

rlucioni:
Looks good @schenedx. I made a few more minor comments. :+1: once they're addressed.

acceptance_tests/pages.py
schenedx: I think you are right. Let me make the change

PR #108, opened 2016-03-17
Add support for program banner images

jimabramson:
work in progress

### DIFF
--- a/programs/apps/api/v1/tests/test_views.py
+++ b/programs/apps/api/v1/tests/test_views.py
@@ -1,14 +1,17 @@
 """
 Tests for Programs API views (v1).
 """
+from cStringIO import StringIO
 import datetime
 import itertools
 import json
 
 import ddt
 from django.core.urlresolvers import reverse
-from django.test import TestCase
+from django.core.files.uploadedfile import SimpleUploadedFile
+from django.test import override_settings, TestCase
 from mock import ANY
+from PIL import Image
 import pytz
 
 from programs.apps.api.v1.tests.mixins import AuthClientMixin, JwtMixin
@@ -188,6 +191,7 @@ class ProgramsViewTests(JwtMixin, TestCase):
                 u"created": ANY,
                 u"modified": ANY,
                 u'marketing_slug': '',
+                u'banner_image_urls': {}
             }
         )
 
@@ -287,6 +291,7 @@ class ProgramsViewTests(JwtMixin, TestCase):
                     u"created": program.created.strftime(DRF_DATE_FORMAT),
                     u"modified": program.modified.strftime(DRF_DATE_FORMAT),
                     u'marketing_slug': program.marketing_slug,
+                    u'banner_image_urls': {}
                 }
             )
 
@@ -314,8 +319,52 @@ class ProgramsViewTests(JwtMixin, TestCase):
                     u"created": program.created.strftime(DRF_DATE_FORMAT),
                     u"modified": program.modified.strftime(DRF_DATE_FORMAT),
                     u'marketing_slug': program.marketing_slug,
+                    u'banner_image_urls': {}
                 }
             )
+
+    def make_banner_image_file(self, name):
+        """
+        Helper to generate values for Program.banner_image
+        """
+        image = Image.new('RGB', (1440, 900), 'green')
+        sio = StringIO()
+        image.save(sio, format='JPEG')
+        return SimpleUploadedFile(name, sio.getvalue(), content_type='image/jpeg')
+
+    def assert_correct_banner_image_urls(self, url_prepend=''):
+        """
+        DRY test helper.  Ensure that the serializer generates a complete set
+        of absolute URLs for the banner_image when one is set.
+        """
+        program = ProgramFactory.create(status=ProgramStatus.ACTIVE)
+        program.banner_image = self.make_banner_image_file('test_filename.jpg')
+        program.save()
+
+        response = self._make_request(program_id=program.id)
+        self.assertEqual(response.status_code, 200)
+
+        expected_urls = {
+            '{}x{}'.format(*size): '{}{}__{}x{}.jpg'.format(url_prepend, program.banner_image.url, *size)
+            for size in program.banner_image.field.sizes
+        }
+        self.assertEqual(response.data[u'banner_image_urls'], expected_urls)
+
+    @override_settings(MEDIA_URL='/test/media/url/')
+    def test_banner_image_urls(self):
+        """
+        Ensure that the request is used to generate absolute URLs for banner
+        images, when MEDIA_ROOT does not specify an explicit host.
+        """
+        self.assert_correct_banner_image_urls(url_prepend='http://testserver')
+
+    @override_settings(MEDIA_URL='https://example.com/test/media/url/')
+    def test_banner_image_urls_with_absolute_media_url(self):
+        """
+        Ensure that banner image URLs are correctly presented when storage
+        is configured to use absolute URLs.
+        """
+        self.assert_correct_banner_image_urls()
 
     def test_view_with_nested(self):
         """
@@ -380,6 +429,7 @@ class ProgramsViewTests(JwtMixin, TestCase):
                 u"created": program.created.strftime(DRF_DATE_FORMAT),
                 u"modified": program.modified.strftime(DRF_DATE_FORMAT),
                 u'marketing_slug': program.marketing_slug,
+                u'banner_image_urls': {}
             }
         )
 
--- a/programs/apps/core/tests/test_s3utils.py
+++ b/programs/apps/core/tests/test_s3utils.py
@@ -1,0 +1,23 @@
+"""
+Test s3 utilities
+"""
+from django.test import TestCase
+from django.conf import settings
+from storages.backends.s3boto import S3BotoStorage
+
+from programs.apps.core.s3utils import MediaS3BotoStorage
+
+
+class MediaS3BotoStorageTestCase(TestCase):
+    """
+    Test the MediaS3BotoStorage django storage driver
+    """
+
+    def test_storage_init(self):
+        """
+        The object is just a partial to S3BotoStorage from django-storages,
+        with some settings piped in.  Ensure this works as expected.
+        """
+        storage = MediaS3BotoStorage()
+        self.assertIsInstance(storage, S3BotoStorage)
+        self.assertEqual(storage.location, settings.MEDIA_ROOT.strip('/'))
--- a/programs/apps/programs/fields.py
+++ b/programs/apps/programs/fields.py
@@ -1,0 +1,206 @@
+"""
+Custom fields used in models in the programs django app.
+"""
+from contextlib import closing
+
+from django.db import models
+from django.db.models.fields.files import ImageFieldFile
+from PIL import Image
+
+from .image_helpers import (
+    create_image_file,
+    crop_image_to_aspect_ratio,
+    scale_image,
+    set_color_mode_to_rgb,
+    validate_image_size,
+    validate_image_type,
+)
+
+
+class ResizingImageFieldFile(ImageFieldFile):
+    """
+    Custom field value behavior for images stored in `ResizingImageField`
+    fields on model instances.  See `ResizingImageField` docs for more info.
+    """
+    @property
+    def resized_names(self):
+        """
+        Return the names of the resized copies of this image (if any), in a
+        dictionary keyed by tuples of (width, height).
+
+        Returns:
+            dict
+        """
+        if not self.name:
+            return {}
+        else:
+            return {
+                (width, height): '{}__{}x{}.jpg'.format(self.name, width, height)
+                for width, height in self.field.sizes
+            }
+
+    @property
+    def resized_urls(self):
+        """
+        Return the URLs of the resized copies of this image (if any), in a
+        dictionary keyed by tuples of (width, height).
+
+        Returns:
+            dict
+        """
+        if not self.name:
+            return {}
+        else:
+            return {size: self.storage.url(name) for size, name in self.resized_names.items()}
+
+    @property
+    def minimum_original_size(self):
+        """
+        Return the minimum acceptable width and height of an uploaded image
+        (which is the same as the greatest (width, height) pair in
+        self.field.sizes).
+
+        Returns:
+            tuple(int, int)
+        """
+        return sorted(self.field.sizes)[-1]
+
+    def create_resized_copies(self):
+        """
+        Generate and store resized copies of the original image, using the
+        django storage API.
+
+        Returns:
+            None
+        """
+        original = Image.open(self.file)
+        image = set_color_mode_to_rgb(original)
+        ref_width, ref_height = self.minimum_original_size
+        image = crop_image_to_aspect_ratio(image, float(ref_width) / float(ref_height))
+
+        for size, name in self.resized_names.items():
+            scaled = scale_image(image, *size)
+            with closing(create_image_file(scaled)) as scaled_image_file:
+                self.storage.save(name, scaled_image_file)
+
+
+class ResizingImageField(models.ImageField):
+    """
+    Customized ImageField that automatically generates and stores a set of
+    resized copies along with the original image files.
+
+    WARNING: this does not presently correct for orientation - processed images
+    taken directly from digital cameras may appear with unexpected rotation.
+
+    TODO: purge stale copies.
+    """
+    attr_class = ResizingImageFieldFile
+
+    def __init__(self, path_template, sizes, *a, **kw):
+        """
+        Arguments:
+
+            path_template (basestring):
+                A format string that will be templated against model
+                instances to produce a directory name for stored files.
+                For example, if your model has a unique "name" field, you
+                could use '/mymodel/{name}/' as the path template.
+
+                To facilitate management/cleanup of stale copies, it's
+                important to use a template that will result in a unique and
+                immutable value for each model object.
+
+                Note that using the primary key ('id') in your template is
+                dangerous, however, because this will evaluate to `None` when
+                initially storing a new model instance which has not yet been
+                assigned an id by the database.  Therefore, choose a value or
+                values which can be assigned before the object is physically
+                saved, for example a UUID or an application-generated timestamp.
+
+            sizes:
+                A sequence of tuples of (width, height) at which to resize
+                copies of the original image.
+
+                The largest of the sizes will be used as the minimum allowed
+                dimensions of a newly-stored file.
+
+                WARNING: presently, all of the sizes must have the same aspect
+                ratio.
+        """
+        if callable(kw.get('upload_to')):
+            # if an upload_to kwarg is passed with a callable value, the
+            # superclass will use it to overwrite the value of
+            # self.generate_filename (which is redefined below).
+            # Since that will lead to unexpected behavior, prevent it from
+            # happening.
+            raise Exception(
+                'ResizingImageField does not support passing a custom callable '
+                'for the `upload_to` keyword arg.'
+            )
+        super(ResizingImageField, self).__init__(*a, **kw)
+        self.path_template = path_template.rstrip('/')
+        self.sizes = sizes
+
+    def generate_filename(self, model_instance, filename):  # pylint: disable=method-hidden
+        """
+        Join our path template with the filename assigned by django storage to
+        generate a filename for newly-uploaded file.
+
+        Arguments:
+
+            model_instance (Model):
+                The model instance whose value is about to be saved.
+
+            filename (basestring):
+                The filename assigned to a newly uploaded file by django.
+
+        Returns:
+
+            ResizingImageFieldFile
+
+        """
+        pathname = self.path_template.format(**model_instance.__dict__)  # pylint: disable=no-member
+        return '{}/{}'.format(pathname, filename)
+
+    def pre_save(self, model_instance, add):
+        """
+        Override pre_save to create resized copies of the original upload
+        when necessary (i.e. a newly stored file).
+
+        Arguments:
+
+            model_instance (Model):
+                The model instance whose value is about to be saved.
+
+            add (bool):
+                Whether the model instance is being added (inserted) for the
+                first time.
+
+        Returns:
+
+            ResizingImageFieldFile
+
+        """
+        # before invoking super, determine if we are dealing with a file that has previously been saved to storage.
+        # we have to check this before calling super since that will store a new file and set _committed to True.
+        original_field_value = getattr(model_instance, self.attname)
+        originally_committed = getattr(original_field_value, '_committed', False)
+
+        field_value = super(ResizingImageField, self).pre_save(model_instance, add)
+
+        # if we just stored a new file, do additional validation, then generate and save resized copies.
+        if not originally_committed:
+            validate_image_type(field_value.file)
+            validate_image_size(field_value.file, *field_value.minimum_original_size)
+            field_value.create_resized_copies()
+
+        return field_value
+
+    def deconstruct(self):
+        """
+        Provide instantiation metadata for the migrations framework.
+        """
+        name, path, args, kwargs = super(ResizingImageField, self).deconstruct()
+        kwargs['sizes'] = self.sizes
+        kwargs['path_template'] = self.path_template
+        return name, path, args, kwargs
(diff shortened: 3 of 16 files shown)


jimabramson:
@rlucioni @schenedx  This is now ready for review.  I regret the size of it, but it was hard to break this change into smaller pieces that made sense independently.

Main areas for review: 
* programs/apps/programs/fields.py along with image_helpers.py and the associated test modules.  These define a reusable auto-resizing image field for use in django models, based in part on code copied from edx-platform.
* programs/apps/programs/models.py and the change in programs/apps/api/serializers.py (and tests).  This is what adds the resizing image field to the program model and API representation.

Cleanup of stale / old resized images will be added in a future PR.

requirements/base.txt
```diff
@@ -13,5 +15,7 @@ edx-rest-api-client==1.5.0
 # Pinning to 0.10.0 is a temporary workaround, but the underlying violation(s) should be fixed.
 libsass==0.10.0
 Markdown==2.6.2
+piexif==1.0.3
+Pillow==3.1.1
```
schenedx: @jimabramson Are you sure the programs IDA stage and production environment support pillow version 3.1.1?
I know that edx-platform only needed Pillow version 2.7.0

Thanks

schenedx:
I wish I can see a demo of this and review again. I feel like I don't have enough knowledge in the area to do a good critic on the change here.

programs/apps/programs/fields.py
```diff
@@ -0,0 +1,192 @@
...
+    set_color_mode_to_rgb,
+    validate_image_size,
+    validate_image_type,
+)
+
+
+class ResizingImageFieldFile(ImageFieldFile):
+    """
+    Custom field value behavior for images stored in `ResizingImageField`
+    fields on model instances.  See `ResizingImageFiled` docs for more info.
```
rlucioni: s/ResizingImageFiled/ResizingImageField

programs/apps/programs/fields.py
```diff
@@ -0,0 +1,192 @@
...
+    taken directly from digital cameras may appear with unexpected rotation.
+
+    TODO: purge stale copies.
+    """
+    attr_class = ResizingImageFieldFile
+
+    def __init__(self, path_template, sizes, *a, **kw):
+        """
+        NOTE: passing a keyword argument to `upload_to` will break this
+        implementation, so don't.
```
rlucioni: This is scary! Why will passing `upload_to` break this? Is it because that path will override the path constructed in `resized_names`?

programs/apps/programs/fields.py
```diff
@@ -0,0 +1,192 @@
...
+                copies of the original image.
+
+                The largest of the sizes will be used as the minimum allowed
+                dimensions of a newly-stored file.
+
+                WARNING: presently, all of the sizes must have the same aspect
+                ratio.
+        """
+        super(ResizingImageField, self).__init__(*a, **kw)
+        self.path_template = path_template
```
rlucioni: Would it make sense to `rstrip('/')` here to make it more likely that this path is well-formed?

programs/apps/programs/fields.py
```diff
@@ -0,0 +1,192 @@
...
+        """
+        # before invoking super, determine if we are dealing with a file that has previously been saved to storage.
+        # we have to check this before calling super since that will store a new file and set _committed to True.
+        original_field_value = getattr(model_instance, self.attname)
+        originally_committed = getattr(original_field_value, '_committed', False)
+
+        field_value = super(ResizingImageField, self).pre_save(model_instance, add)
+
+        # if we just stored a new file, do additional validation, then generate and save resized copies.
+        if field_value.name and not originally_committed:
```
rlucioni: Is it possible for `field_value.name` to come back as `None` here? Do we need to do something special if that's the case? Is checking `originally_committed` enough?

programs/apps/programs/image_helpers.py
```diff
@@ -0,0 +1,210 @@
...
+    pass
+
+
+ImageType = namedtuple('ImageType', ('extensions', 'mimetypes', 'magic'))
+
+IMAGE_TYPES = {
+    'jpeg': ImageType(
+        extensions=['.jpeg', '.jpg'],
+        mimetypes=['image/jpeg', 'image/pjpeg'],
+        magic=['ffd8'],
```
rlucioni: It'd be nice to demystify this by calling the field `file_signatures`. A link to https://en.wikipedia.org/wiki/List_of_file_signatures might not hurt, either.

programs/apps/programs/image_helpers.py
```diff
@@ -0,0 +1,210 @@
...
+    """
+    # validation code by @pmitros,
+    # adapted from https://github.com/pmitros/ProfileXBlock
+    # see also: http://en.wikipedia.org/wiki/Magic_number_%28programming%29
+
+    uploaded_file.seek(0)
+
+    # check the file extension looks acceptable
+    filename = unicode(uploaded_file.name).lower()
+    filetype = [ft for ft in IMAGE_TYPES if any(filename.endswith(ext) for ext in IMAGE_TYPES[ft].extensions)]
```
rlucioni: Would be a bit more readable as
```python
filetype = [filetype for filetype, imagetype in IMAGE_TYPES.items() if any(filename.endswith(ext) for ext in imagetype.extensions)]
```

programs/apps/programs/image_helpers.py
```diff
@@ -0,0 +1,210 @@
...
+    string_io = StringIO()
+    image.save(string_io, format='JPEG')
+    return ContentFile(string_io.getvalue())
+
+
+def _get_valid_file_types():
+    """
+    Return comma separated string of valid file types.
+    """
+    return ', '.join([', '.join(IMAGE_TYPES[ft].extensions) for ft in IMAGE_TYPES.keys()])
```
rlucioni: It's not necessary to call `keys()` here.

programs/apps/programs/image_helpers.py
```diff
@@ -0,0 +1,210 @@
...
+
+    # check the file extension looks acceptable
+    filename = unicode(uploaded_file.name).lower()
+    filetype = [ft for ft in IMAGE_TYPES if any(filename.endswith(ext) for ext in IMAGE_TYPES[ft].extensions)]
+    if not filetype:
+        file_upload_bad_type = _(
+            u'The file must be one of the following types: {valid_file_types}.'
+        ).format(valid_file_types=_get_valid_file_types())
+        raise ImageValidationError(file_upload_bad_type)
+    filetype = filetype[0]
```
rlucioni: What we actually need at this point is the image type. What about assigning:
```python
image_type = IMAGE_TYPES[filetypes[0]]
```

programs/apps/programs/image_helpers.py
```diff
@@ -0,0 +1,210 @@
...
+    Returns:
+        None
+
+    Raises:
+        ImageValidationError:
+            when the image file is an unsupported or invalid type.
+    """
+    # validation code by @pmitros,
+    # adapted from https://github.com/pmitros/ProfileXBlock
+    # see also: http://en.wikipedia.org/wiki/Magic_number_%28programming%29
```
rlucioni: Nitpick: whatever we want to keep here looks like it belongs in the docstring.

programs/apps/programs/image_helpers.py
```diff
@@ -0,0 +1,210 @@
...
+        minimum_height (int): minimum height of the image in pixels
+
+    Returns:
+        None
+
+    Raises:
+        ImageValidationError:
+            when the image file is an unsupported or invalid type.
+    """
+    image_width, image_height = Image.open(uploaded_file).size
```
rlucioni: Should the file be opened with a context manager so that it's closed when we're done with it?
```python
with Image.open(uploaded_file) as image:
    image_width, image_height = image.size
```

programs/apps/programs/fields.py
```diff
@@ -0,0 +1,192 @@
...
+
+    def create_resized_copies(self):
+        """
+        Generate and store resized copies of the original image, using the
+        django storage API.
+
+        Returns:
+            None
+        """
+        original = Image.open(self.file)
```
rlucioni: Should the file be opened with a context manager so that the file pointer is closed when we're done with it?

programs/apps/programs/image_helpers.py
```diff
@@ -0,0 +1,210 @@
...
+    in memory (not on disk).
+
+    Arguments:
+        image (Image)
+
+    Returns:
+        ContentFile
+    """
+    string_io = StringIO()
+    image.save(string_io, format='JPEG')
```
rlucioni: I don't understand what's being done here. Could you explain?

programs/apps/programs/models.py
```diff
@@ -63,6 +72,13 @@ class Program(TimeStampedModel):
         max_length=255
     )
 
+    banner_image = ResizingImageField(
+        path_template='program/banner/{uuid}',
+        sizes=[(1440, 900), (480, 300), (360, 225), (200, 125)],
```
rlucioni: @schenedx I remember there was talk about which banner image sizes would be available. Are these what you were expecting?

programs/settings/base.py
```diff
@@ -284,3 +284,13 @@
...
+
+
+# OUR PUBLIC URL
+# scheme + host (+ port); no trailing slash.
+#
+# Note: this is used in production-like environments to ensure URLs are
+# generated using a specific host/domain instead of using the HTTP request
+# information (since the request may have been made over a private network).
+# In development environments it is OK to leave empty.
+PUBLIC_URL_ROOT = ''
```
rlucioni: I don't see this setting used anywhere. Am I missing something?

programs/settings/production.py
```diff
@@ -21,11 +21,18 @@
 
 LOGGING = get_logger_config()
 
+# This may be overridden by the yaml in PROGRAMS_CFG, but it should
+# be here as a default.
```
rlucioni: In that case, put it in the base configuration?

requirements/base.txt
rlucioni: @jimabramson we need to add libjpeg to the programs role's Debian package list in order to use Pillow's JPEG functionality. Here's how we did it for [ecommerce](https://github.com/edx/configuration/blob/master/playbooks/roles/ecommerce/defaults/main.yml#L210).

programs/apps/programs/models.py
schenedx: @rlucioni Yes, these are. Thanks for checking in.

programs/apps/api/v1/tests/test_views.py
```diff
@@ -314,9 +319,53 @@ def test_view_learner(self, status):
...
+        }
+        self.assertEqual(response.data[u'banner_image_urls'], expected_urls)
+
+    @override_settings(MEDIA_URL='/test/media/url/')
+    def test_banner_image_urls(self):
+        """
+        Ensure that the request is used to generate absolute URLs for banner
+        images, when MEDIA_ROOT does not specify an explicit host.
+        """
+        self.assert_correct_banner_image_urls('http://testserver')
```
rlucioni: Nitpick: for clarity, consider using your keyword argument as a keyword argument:
```python
self.assert_correct_banner_image_urls(url_prepend='http://testserver')
```

programs/apps/core/tests/test_s3utils.py
```diff
@@ -0,0 +1,22 @@
+"""
+Test s3 utilities
+"""
+from django.test import TestCase
+from django.conf import settings
+
+from programs.apps.core.s3utils import MediaS3BotoStorage, S3BotoStorage
```
rlucioni: Could you import `S3BotoStorage` directly from `storages.backends.s3boto` instead of another module which happens to import it?

programs/apps/programs/tests/helpers.py
```diff
@@ -0,0 +1,62 @@
...
+import os
+from tempfile import NamedTemporaryFile
+
+from django.core.files.uploadedfile import UploadedFile
+import piexif
+from PIL import Image
+
+
+@contextmanager
+def make_image_file(dimensions=(320, 240), extension=".jpeg", force_size=None, orientation=None):
```
rlucioni: Would it make sense to roll these context managers into an `ImageFileTestMixin`?

programs/apps/programs/tests/helpers.py
```diff
@@ -0,0 +1,62 @@
+"""
+Helper methods for use in profile image tests.
```
rlucioni: "Helper methods for use in ~~profile~~ image tests."

programs/apps/programs/tests/test_fields.py
```diff
@@ -0,0 +1,156 @@
...
+    )
+    @ddt.unpack
+    def test_pre_save(self, filename, is_existing_file, mock_validate_size, mock_validate_type):
+        """
+        Ensure that image validation and resizing take place only when a new
+        file is being stored.
+        """
+        # pylint: disable=too-many-function-args
+
+        # create field_value, and wire up to test model + field instances
```
rlucioni: I think you can get rid of this comment.

programs/apps/programs/tests/test_fields.py
```diff
@@ -0,0 +1,156 @@
...
+            with make_uploaded_file('image/jpeg', (1000, 1000)) as image_file:
+                if filename:
+                    field_value.file = image_file
+                    field_value._committed = is_existing_file  # pylint: disable=protected-access
+                self.field.pre_save(self.model_instance, False)
+
+        expected_called = bool(filename) and not is_existing_file
+        self.assertEqual(mock_validate_size.called, expected_called)
+        self.assertEqual(mock_validate_type.called, expected_called)
+        self.assertEqual(mock_resize.called, expected_called)
```
rlucioni: Consider DRYing this out:
```python
for actual_called in (mock_validate_size.called, mock_validate_type.called, mock_resize.called):
    self.assertEqual(actual_called, expected_called)
```

programs/apps/programs/tests/test_image_helpers.py
```diff
@@ -0,0 +1,162 @@
+"""
+Test cases for image processing functions in the profile image package.
```
rlucioni: Scratch "profile"?

programs/apps/api/v1/tests/test_views.py
```diff
@@ -314,9 +319,53 @@ def test_view_learner(self, status):
...
                     u'marketing_slug': program.marketing_slug,
+                    u'banner_image_urls': {}
                 }
             )
 
+    def make_banner_image_file(self, name):
+        """
+        Helper to generate values for Program.banner_image
+        """
+        image = Image.new('RGB', (1440, 900), "green")
```
rlucioni: Nitpick: mixed single and double quotes. Could we stick to singles?

rlucioni:
@jimabramson I've completed my first pass.

requirements/base.txt
jimabramson: @rlucioni Thanks for the heads up - see https://github.com/edx/configuration/pull/2883.
@schenedx The requirement has been satisfied on a sandbox which suggests to me that whatever dependencies we are installing for other roles are sufficient to support 3.1.1 - let me know if you see a reason not to use the most recent version here.

programs/apps/programs/fields.py
jimabramson: I'm adding explicit handling and some inline comments to explain this better.

programs/apps/programs/fields.py
jimabramson: good call

programs/apps/programs/fields.py
jimabramson: I thought it was possible, but apparently it is not, so I have removed it.

programs/apps/programs/models.py
jimabramson: Actually @schenedx @rlucioni that has changed as of @marcotuts' most recent comment here

https://openedx.atlassian.net/browse/ECOM-3196?focusedCommentId=178035&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-178035

So we need to update these.   The new aspect ratio is 3:1 - most of the widths will work but perhaps we should bump the 200px width down to 180px; @schenedx what do you think?

programs/settings/base.py
jimabramson: As it turns out, actually no.  This was added in order to support generating absolute URLs for resized banner images, but I realize the proper way to do that is via specifying an absolute URL base in `MEDIA_URL`.  So will remove.

programs/settings/production.py
jimabramson: Not sure.  This pattern is based on one used with the credentials IDA.  `MEDIA_STORAGE_BACKEND` is not a setting of its own (never referred to in code), but a device specific to the .yml file for grouping the related individual storage settings.  So I think it's specific to this place here.

programs/apps/api/v1/tests/test_views.py
jimabramson: agree

programs/apps/core/tests/test_s3utils.py
jimabramson: done

programs/apps/programs/tests/test_fields.py
jimabramson: That's deliberately there to help the next maintainer of this code understand what is going on, but if you think it's self-evident, I can get rid of it.

programs/apps/programs/image_helpers.py
jimabramson: in order to turn a PIL Image object into a file-like object (so that django storage APIs can work with it), we want an in-memory filelike object to which the PIL Image.save call can write data.  ContentFile is a good destination for the data but we can't Image.save directly to it; instead we use the StringIO buffer as an intermediary and initialize the ContentFile from its value.

rlucioni: Got it, thank you.

programs/settings/production.py
rlucioni: That makes sense.

programs/apps/programs/tests/test_fields.py
rlucioni: This one in particular struck me as self-evident.

programs/apps/programs/image_helpers.py
jimabramson: Calling `Image.close()` would have the undesirable side-effect of closing the underlying file (see [docs](http://pillow.readthedocs.org/en/3.1.x/reference/Image.html#PIL.Image.Image.close))

programs/apps/programs/fields.py
jimabramson: same here, `Image.close()` should only be called when there's a desire to close whatever file handle was used to initialize it, which isn't desired here.

programs/apps/programs/tests/helpers.py
jimabramson: It would work but I'm not sure what the benefit would be.  The context managers don't need any references to `self` and therefore don't get any benefit from binding to object instances, so I don't think they'd be any more reusable than they are now.  Do you see other potential advantages?

programs/apps/programs/tests/test_fields.py
jimabramson: poof! it's gone

programs/apps/programs/tests/helpers.py
rlucioni: No, you make a good point. There doesn't seem to be a benefit from binding to object instances here at this point.

jimabramson:
@schenedx let me know what kind of demonstration would help facilitate your review.  In practice there's not a ton to look at, except the additional field exposed in django admin and the URLs added to the program API object, but ping if you want a deeper walkthrough.

programs/apps/programs/models.py
```diff
@@ -63,6 +72,13 @@ class Program(TimeStampedModel):
         max_length=255
     )
 
+    banner_image = ResizingImageField(
+        path_template='program/banner/{uuid}',
+        sizes=[(1440, 480), (480, 160), (360, 120), (180, 60)],
```
rlucioni: @jimabramson what happens if we want to change these sizes or add an additional one?

rlucioni:
@jimabramson looking good, I just have one more question.

jimabramson:
@cpennington per discussion this morning - feedback appreciated.

programs/apps/programs/image_helpers.py
schenedx: @jimabramson I also benefited from your explanation above, can you put the explanation in code as comments for others?

programs/apps/programs/models.py
```diff
@@ -23,6 +26,12 @@ class Program(TimeStampedModel):
     Representation of a Program.
     """
 
+    uuid = models.UUIDField(
```
schenedx: Why do we need an UUID for the program in this PR?

schenedx:
@jimabramson Besides the minor request and question I made, rest LGTM. :+1:

programs/apps/programs/models.py
jimabramson: @rlucioni adding / removing sizes is a code change here; django will want a migration to be generated (though it will not actually touch the database, so editing the existing migration in-place will be a safe option).

To regenerate copies after changing the set of sizes, one low-tech option would be to download and re-upload the original image files.  A management command could be created to automate that process (without actually having to re-transfer any files); I haven't implemented support for this here but I think it will make sense to do this if and when we have a need for it.

programs/apps/programs/models.py
jimabramson: @schenedx good question; i'm glad you called this out.  This arose out of the need to have a naming scheme for the stored files that can be programmatically generated based on the identity of the program.

Initially, the idea was `'program/banner/{id}'` but the problem is that when associating a file during creation, the image files will be stored before the record has been saved to the database and assigned a permanent `id`.

Generating a UUID for each program in the application solves this; it's a secondary unique id that we know ahead of time, so we can use it to construct a reliable two-way navigation between a program in the database and the files on "disk" which will facilitate synchronization and/or cleanup operations on files.

jimabramson: btw now that we have a UUID we may choose to expose this in course discovery or other public APIs rather than the sequential database-assigned IDs.  Since Programs are not very high cardinality I'm not sure how useful that is, but it's an option.   cc: @clintonb @bderusha @cpennington

programs/apps/programs/models.py
rlucioni: Got it, thank you.

rlucioni:
:+1:

programs/apps/programs/models.py
schenedx: Got it. Agreed

programs/apps/programs/fields.py
```diff
@@ -0,0 +1,199 @@
...
+    def minimum_original_size(self):
+        """
+        Return the minimum acceptable width and height of an uploaded image
+        (which is the same as the greatest (width, height) pair in
+        self.field.sizes).
+
+        Returns:
+            tuple(int, int)
+        """
+        return sorted(self.field.sizes)[-1]
```
cpennington: This will sort by width first, and then tie-break by height. Is that ok?

jimabramson: @cpennington yes.  a present limitation of the field is that the aspect ratio must be the same for all dimensions, so a non-deterministic result should never happen except by developer error.

programs/apps/programs/image_helpers.py
```diff
@@ -0,0 +1,216 @@
+"""
+Image file manipulation functions.
+
+This module was based heavily on
+edx-platform/openedx/core/djangoapps/profile_images/images.py, but needed
+further generalization to provide correct functionality in this app.
```
cpennington: Should we be moving this stuff into a repo that can be used by platform and programs (and future projects that need image stuff?)

cpennington:
:+1:

jimabramson:
@schenedx @rlucioni @cpennington FYI i've pushed a commit adding some details to inline comments as discussed herein.

programs/apps/programs/image_helpers.py
jimabramson: @cpennington yes, i think that makes sense.  Factoring a common library out of stuff needed/used by platform and programs would have significantly expanded the scope of this story though, so I pulled back from that.  Let's follow up offline and see where that work might fit.

schenedx:
:shipit:

PR #109, opened 2016-03-21
Install opaque-keys from PyPI

rlucioni:
The opaque-keys package is available on PyPI. FYI @schenedx @jimabramson @clintonb


clintonb:
:+1:

PR #115, opened 2016-04-05
ECOM-3198 Add a new set of sizes to the banner image for display

schenedx:
The original size does not fit the dimensions of the program cards well.
Tweak the sizes of the banner images so the program cards looks better


jimabramson:
@schenedx should be fine 👍  (after rebase, and confirming with @marcotuts that these are the final sizes)

PR #116, opened 2016-04-06
Revert "Upgrade to Django 1.9.4"

schenedx:
This reverts commit 59764104a6ae8bf1d6877aed825696f89c7394d9.


jimabramson:
👍

nedbat:
Just noticed this PR.  There is a long-standing open issue about whether we should aim for LTS releases where we can, or whether we should upgrade eagerly.  Somehow we should get this raised.

jimabramson:
@nedbat excellent point, thanks for bringing that up.

jimabramson:
follow up is in ARCH-39

PR #118, opened 2016-04-08
ECOM-3198 add the new size for banner images for program listings page

schenedx:
Add one more size to process for programs banner images
Tested:
Make sure on the program listing page, the cards are responsive and the images displays correctly for each view port size break points


programs/apps/programs/models.py
```diff
@@ -73,7 +73,7 @@ class Program(TimeStampedModel):
 
     banner_image = ResizingImageField(
         path_template='program/banner/{uuid}',
-        sizes=[(1440, 480), (726, 242), (435, 145), (348, 116)],
+        sizes=[(1440, 480), (726, 242), (498, 166), (435, 145), (348, 116)],
```
AlasdairSwan: I am a little concerned about the number of images we are handling here. What is the file size difference between 498x166 and 435x145? I'm not sold on the idea that we need 5 different images for the program cards at various screenwidths.

schenedx:
@AlasdairSwan helped me with an CSS trick that fixed the left blank issue with banner image without needing this change in the PR. Closing it

PR #121, opened 2016-04-19
Set max_length to 1000 for banner image filenames.

jimabramson:
Ensure the filenames do not exceed [the maximum length of a key in s3](http://docs.aws.amazon.com/AmazonS3/latest/dev/UsingMetadata.html).

1000 is used instead of 1024, which allows some headroom for the names of the resized copies (which have ~12 characters added to the original name).

@schenedx @rlucioni PTAL.


rlucioni:
👍

schenedx:
👍

PR #122, opened 2016-04-19
Update Travis config

rlucioni:
An issue preventing Travis from caching the pip directory (travis-ci/travis-ci#3239) has been resolved. It's also a good idea to run builds for both commits and merge commits.

Counterpart to https://github.com/edx/cookiecutter-django-ida/pull/23. @clintonb


clintonb:
👍

PR #126, opened 2016-04-27
Fix overly specific get_or_create preventing org sync

rlucioni:
In case their display name has changed, orgs should initially be looked up by key. The display name can be updated separately.

@schenedx or @jimabramson, could I get a review from one of you? @jibsheet FYI, this will fix the error we observed.

### DIFF
--- a/programs/apps/programs/management/commands/sync_orgs.py
+++ b/programs/apps/programs/management/commands/sync_orgs.py
@@ -52,14 +52,16 @@ class Command(BaseCommand):
 
             for org in orgs:
                 if org['active']:
-                    fields = {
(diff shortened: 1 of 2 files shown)


schenedx:
@rlucioni LGTM, although I am not familiar with this area. @jimabramson could you please help give another review?

rlucioni:
@schenedx this is a management command used to sync org information from the LMS. I'm going to go ahead and merge this. @jimabramson if you get a chance to look and see something glaring, please let me know and I'll revise as necessary.

jimabramson:
@rlucioni good call. 👍

PR #128, opened 2016-04-28
ECOM-3950 clean up stale copies of program banner images

schenedx:
@rlucioni @jimabramson Please review
This is the AC#1 of the ticket
Tested on django admin:
1. Upload a new banner image. Make sure it succeeds
2. Upload a banner image on top of existing banner image, make sure it succeeds and both images set are kept
3. Upload banner images on the same program multiple times and make sure the save succeeds and only 2 set of most recent images are kept


programs/apps/programs/fields.py
```diff
@@ -83,6 +83,49 @@ def create_resized_copies(self):
...
+        if self.name:
+            if self.name in files_to_clean:
+                files_to_clean.remove(self.name)
+            for _, name in self.resized_names.items():
+                if name in files_to_clean:
+                    files_to_clean.remove(name)
+
+        files_to_clean.sort(key=lambda x: self.storage.modified_time(x))    # pylint: disable=unnecessary-lambda
+        size = len(files_to_clean)
+        if size > len(self.field.sizes) + 1:
```
jimabramson: One thing that concerns me about this approach is that it assumes any stray files were parts of a complete set and that the number of resizes images has never changed.  I think eventually this assumption will break.

It does change the approach somewhat, but I think an approach that groups files by a common original filename, and assigns a date to that group (e.g. max modified time across files in that group), would be more future-proof.

programs/apps/programs/fields.py
```diff
@@ -83,6 +83,49 @@ def create_resized_copies(self):
...
+            # Leave the most recent set before the save, in case we need to roll back
+            # The extra 1 is for the orginal file saved before resizing
+            files_to_clean = files_to_clean[0:size - len(self.field.sizes) - 1]
+            for file_to_clean in files_to_clean:
+                self.storage.delete(file_to_clean)
```
jimabramson: i would advocate for some logging here which we may appreciate if we ever run into unexpectedly removed (or unexpectedly piled-up) images and need to figure out what happened.

programs/apps/programs/fields.py
```diff
@@ -83,6 +83,49 @@ def create_resized_copies(self):
...
+            return
+        # Remove existing image files
+        if self.name:
+            if self.name in files_to_clean:
+                files_to_clean.remove(self.name)
+            for _, name in self.resized_names.items():
+                if name in files_to_clean:
+                    files_to_clean.remove(name)
+
+        files_to_clean.sort(key=lambda x: self.storage.modified_time(x))    # pylint: disable=unnecessary-lambda
```
jimabramson: just as a heads up, I believe this has the potential to be expensive if made over a large number of files in s3, because there will be a separate s3 API call for each.   More than a few files is not expected under normal conditions, and I don't think it's worth the complexity of writing a custom s3 API call to list the stored keys with their mod times, so this is mainly an advisory point.

jimabramson:
@rlucioni @schenedx I've pushed a proposed revision as a fixup commit.

programs/apps/programs/fields.py
```diff
@@ -83,6 +89,45 @@ def create_resized_copies(self):
...
+
+    def clean_stale_images(self, keep_previous=True):
+        """
+        Search and clean historical images left in the storage,
+        using django storage API
+
+        Returns:
+            None
+        """
+        dir_ = self.field.get_path(self.instance)
```
rlucioni: Why the trailing underscore?

jimabramson: `dir` is reserved (builtin function)

rlucioni:
👍

rlucioni:
@jimabramson let me know if you're unable to squash commits or merge.

programs/apps/programs/fields.py
```diff
@@ -83,6 +89,45 @@ def create_resized_copies(self):
...
+
+        groups = {}
+        for name in self.storage.listdir(dir_)[1]:
+            base_name = re.sub(r'__([\d]+)x([\d]+)\.jpg', '', name)
```
schenedx: This regular expression is my only concern, not a blocker though. I have seen the file name to be something like "headerImage_1400x480_geeFegfdsE.jpg". The time I have seen this file name show up is when the clean up logic was incorrect and left some image in the dir. So, I guess this is OK for now.

jimabramson:
based on manual testing i just did the implementation seems to have a bug.  not merging just yet.

schenedx:
@jimabramson Are you going to help us merge this PR in even tho you are no longer with edX? Thank you very much!

jimabramson:
np - after last fix should be GTG

jimabramson:
or not

jimabramson:
@schenedx @rlucioni i'm happy to squash and merge this if you're cool with the bugfix commit(s) i added.

jimabramson:
or y'all can push the button too :-)

schenedx:
@jimabramson I think this LGTM.
If you got the chance, please Squash and Merge

rlucioni:
Thank you @jimabramson!

PR #129, opened 2016-04-28
add unique constraint to Program.uuid

jimabramson:
ECOM-4305

@rlucioni @schenedx I have a little more local / manual testing to do on the image re-save part, but so far everything is working as desired, and I consider this ready for review.

NOTE the approach is heavily based on [this django documentation](https://github.com/django/django/commit/1f9e44030e9c5300b97ef7b029f482c53a66f13b)


rlucioni:
@jimabramson could you help me understand how this happened in the first place? I thought we were following Django's [model reference](https://docs.djangoproject.com/en/1.8/ref/models/fields/#uuidfield). Is their documentation wrong?

schenedx:
@jimabramson This solution is cool. 👍 
Please make sure this is merged before https://github.com/edx/programs/pull/128

programs/apps/programs/migrations/0009_make_program_uuid_unique.py
```diff
@@ -0,0 +1,55 @@
...
+import os
+import uuid
+
+from django.core.files.uploadedfile import SimpleUploadedFile
+from django.db import migrations, models
+
+
+def regenerate_uuids(apps, schema_editor):
+    """
+    """
```
rlucioni: Missing docstring.

programs/apps/programs/migrations/0009_make_program_uuid_unique.py
```diff
@@ -0,0 +1,55 @@
...
+    Program = apps.get_model('programs', 'Program')
+
+    # first ensure this needs to be done at all.  If no two programs currently
+    # have the same uuid value, there's no reason to modify anything.
+    # NOTE: if ANY programs have non-unique uuids, ALL programs will have their
+    # uuids regenerated.
+    counts_by_uuid = Program.objects.values('uuid').annotate(models.Count('id'))
+    if not any(map(lambda r: r['id__count'] > 1, counts_by_uuid)):
```
rlucioni: This would be more readable as a list comprehension.
```python
any([r['id__count'] > 1 for r in counts_by_uuid])
```

jimabramson:
@rlucioni as far as the missing constraint itself, that was human error (guess who).

As far as why unique values weren't assigned in the first place, it's discussed to some length in these two Django tickets (see below).  To summarize, that a callable default field value is only computed once during a migration is expected behavior, though IMO it's counterintuitive and not very well justified. 

* https://code.djangoproject.com/ticket/23408
* https://code.djangoproject.com/ticket/23932

The relevant bit of doc I linked in the main description of this PR is part of official Django docs - had I not omitted the unique constraint in the first place, our original migration that added the uuid field would have looked similar to what's here.

programs/apps/programs/migrations/0009_make_program_uuid_unique.py
jimabramson: agree, changing.

jimabramson:
@rlucioni updated with requested changes.

rlucioni:
👍

jimabramson:
validated locally and rebased, will merge once tests pass.

rlucioni:
@jimabramson you're good to merge. Codecov is having problems and probably won't come back any time soon: http://status.codecov.io/.

jimabramson:
@rlucioni thanks for the headsup on codecov.

PR #130, opened 2016-04-29
Unpin PSA

rlucioni:
@clintonb


clintonb:
Update `edx-auth-backends`?

rlucioni:
In progress: https://travis-ci.org/edx/auth-backends/builds/126738621.

rlucioni:
#131

PR #132, opened 2016-05-05
ECOM-3237 Default Program Images for Programs Listing Page

schenedx:
@rlucioni Please review

Tested using django admin:

- Created a default program banner image within the DefaultProgramValue object
-- Made sure the API returned the default image
-- Made sure when a program already have the banner image defined, the program defined banner image urls are returned
- Try to add a new DefaultProgramValue object and made sure the error is handled gracefully in django admin

### DIFF
--- a/programs/apps/api/serializers.py
+++ b/programs/apps/api/serializers.py
@@ -269,11 +269,26 @@ class ProgramSerializer(serializers.ModelSerializer):
     organizations = ProgramOrganizationSerializer(many=True, source='programorganization_set')
     course_codes = ProgramCourseCodeSerializer(many=True, source='programcoursecode_set', required=False)
 
+    def _get_default_banner_images(self):
+        """Get default banner image URLs.
+
+        Returns:
+            list of tuples if default banner image has been configured. Empty list otherwise.
+        """
+        try:
+            program_default = models.ProgramDefault.objects.get()
+            return program_default.banner_image.resized_urls.items()
+        except models.ProgramDefault.DoesNotExist:
+            return []
+
     def get_banner_image_urls(self, instance):
         """
         Render public-facing URLs for the available banner images.
         """
         url_items = instance.banner_image.resized_urls.items()
+        if not url_items:
+            url_items = self._get_default_banner_images()
+
         # in case MEDIA_URL does not include scheme+host, ensure that the URLs are absolute and not relative
         url_items = [[size, self.context['request'].build_absolute_uri(url)] for size, url in url_items]
         return {'w{}h{}'.format(*size): url for size, url in url_items}
--- a/programs/apps/api/v1/tests/test_views.py
+++ b/programs/apps/api/v1/tests/test_views.py
@@ -1,17 +1,14 @@
 """
 Tests for Programs API views (v1).
 """
-from cStringIO import StringIO
 import datetime
 import itertools
 import json
 
 import ddt
 from django.core.urlresolvers import reverse
-from django.core.files.uploadedfile import SimpleUploadedFile
 from django.test import override_settings, TestCase
 from mock import ANY
-from PIL import Image
 import pytz
 
 from programs.apps.api.v1.tests.mixins import AuthClientMixin, JwtMixin
@@ -19,6 +16,7 @@ from programs.apps.core.constants import Role
 from programs.apps.core.tests.factories import UserFactory
 from programs.apps.programs.constants import ProgramCategory, ProgramStatus
 from programs.apps.programs.models import CourseCode, Program, ProgramCourseCode, ProgramCourseRunMode
+from programs.apps.programs.tests.helpers import make_banner_image_file
 from programs.apps.programs.tests.factories import (
     CourseCodeFactory,
     OrganizationFactory,
@@ -26,6 +24,7 @@ from programs.apps.programs.tests.factories import (
     ProgramFactory,
     ProgramOrganizationFactory,
     ProgramCourseRunModeFactory,
+    ProgramDefaultFactory
 )
 
 
@@ -323,22 +322,13 @@ class ProgramsViewTests(JwtMixin, TestCase):
                 }
             )
 
-    def make_banner_image_file(self, name):
-        """
-        Helper to generate values for Program.banner_image
-        """
-        image = Image.new('RGB', (1440, 900), 'green')
-        sio = StringIO()
-        image.save(sio, format='JPEG')
-        return SimpleUploadedFile(name, sio.getvalue(), content_type='image/jpeg')
-
     def assert_correct_banner_image_urls(self, url_prepend=''):
         """
         DRY test helper.  Ensure that the serializer generates a complete set
         of absolute URLs for the banner_image when one is set.
         """
         program = ProgramFactory.create(status=ProgramStatus.ACTIVE)
-        program.banner_image = self.make_banner_image_file('test_filename.jpg')
+        program.banner_image = make_banner_image_file('test_filename.jpg')
         program.save()
 
         response = self._make_request(program_id=program.id)
@@ -347,6 +337,38 @@ class ProgramsViewTests(JwtMixin, TestCase):
         expected_urls = {
             'w{}h{}'.format(*size): '{}{}__{}x{}.jpg'.format(url_prepend, program.banner_image.url, *size)
             for size in program.banner_image.field.sizes
+        }
+        self.assertEqual(response.data[u'banner_image_urls'], expected_urls)
+
+    def assert_correct_default_banner_image_urls(self, url_prepend='', set_program_banner=False):
+        """
+        DRY test helper.  Ensure that the serializer generates a complete set
+        of absolute URLs for the default banner_image when the banner image on program is not set.
+        If the flag set_program_banner is defined, then make sure the banner image for program is set
+        and returned
+        """
+        default_program_value = ProgramDefaultFactory.create()
+        default_program_value.banner_image = make_banner_image_file('default_test_filename.jpg')
+        default_program_value.save()
+
+        program = ProgramFactory.create(status=ProgramStatus.ACTIVE)
+        if set_program_banner:
+            program.banner_image = make_banner_image_file('test_filename.jpg')
+            program.save()
+
+        response = self._make_request(program_id=program.id)
+        self.assertEqual(response.status_code, 200)
+
+        banner_image_instance = default_program_value.banner_image
+        if set_program_banner:
+            banner_image_instance = program.banner_image
+
+        expected_urls = {
+            'w{}h{}'.format(*size): '{}{}__{}x{}.jpg'.format(
+                url_prepend,
+                banner_image_instance.url,
+                *size)
+            for size in banner_image_instance.field.sizes
         }
         self.assertEqual(response.data[u'banner_image_urls'], expected_urls)
 
@@ -365,6 +387,29 @@ class ProgramsViewTests(JwtMixin, TestCase):
         is configured to use absolute URLs.
         """
         self.assert_correct_banner_image_urls()
+
+    @override_settings(MEDIA_URL='/test/media/url/')
+    def test_default_banner_image_urls(self):
+        """
+        Ensure that the request is used to generate absolute URLs for the default banner
+        images, when MEDIA_ROOT does not specify an explicit host.
+        """
+        self.assert_correct_default_banner_image_urls(url_prepend='http://testserver')
+
+    @override_settings(MEDIA_URL='https://example.com/test/media/url/')
+    def test_default_banner_image_urls_with_absolute_media_url(self):
+        """
+        Ensure that default banner image URLs are correctly presented when storage
+        is configured to use absolute URLs.
+        """
+        self.assert_correct_default_banner_image_urls()
+
+    @override_settings(MEDIA_URL='https://example.com/test/media/url/')
+    def test_program_banner_image_url_with_default_defined(self):
+        """
+        Ensure the program defined banner image url is returned even if default banner image url is defined
+        """
+        self.assert_correct_default_banner_image_urls(set_program_banner=True)
 
     def test_view_with_nested(self):
         """
--- a/programs/apps/programs/admin.py
+++ b/programs/apps/programs/admin.py
@@ -3,6 +3,7 @@ from django import forms
 from django.contrib import admin
 from opaque_keys import InvalidKeyError
 from opaque_keys.edx.keys import CourseKey
+from solo.admin import SingletonModelAdmin
 
 from programs.apps.programs import models
 from programs.apps.programs.image_helpers import validate_image_type, validate_image_size
@@ -83,6 +84,7 @@ class ProgramCourseRunModeForm(forms.ModelForm):
     """Model form for ProgramCourseCode model. Adding custom validation for
     course key format at form level.
     """
+
     def clean_course_key(self):
         """Clean the course key to make sure format is valid."""
         course_key = self.cleaned_data['course_key']
@@ -117,3 +119,4 @@ admin.site.register(models.ProgramOrganization, ProgramOrganizationAdmin)
 admin.site.register(models.CourseCode, CourseCodeAdmin)
 admin.site.register(models.ProgramCourseCode, ProgramCourseCodeAdmin)
 admin.site.register(models.ProgramCourseRunMode, ProgramCourseRunModeAdmin)
+admin.site.register(models.ProgramDefault, SingletonModelAdmin)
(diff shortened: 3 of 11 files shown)


schenedx:
@clintonb Can I also borrow your time on this review?

schenedx:
@bderusha Can you please help review this PR?

programs/apps/programs/models.py
```diff
@@ -275,3 +275,28 @@ def save(self, *_a, **__kw):
...
 
         return super(ProgramCourseRunMode, self).save()
+
+
+class DefaultProgramValue(TimeStampedModel):
+    """
+    The model to store the default value set for all programs.
+    The image set stored in this model will be the default image presented through the API
+    when program specific image are not specified. This model is custom constructed to
+    only take one 1 instance of the model
```
rlucioni: I'd replace the last sentence with: "Only one instance of this model can be saved."

programs/apps/programs/models.py
```diff
@@ -275,3 +275,28 @@ def save(self, *_a, **__kw):
...
+        null=True,
+        blank=True,
+        max_length=1000,
+    )
+
+    def save(self, *_a, **__kw):
+        """
+        Override save to make sure we only have one such instance active in db
+        """
+        if self.__class__.objects.exists() and self.__class__.objects.all()[0].id != self.id:
```
rlucioni: Couldn't you do this more simply with:
```python
if self.__class__.objects.all().count() > 0:
    raise ValidationError(_("Cannot save more than 1 instance of this model"))
```

programs/apps/programs/models.py
```diff
@@ -275,3 +275,28 @@ def save(self, *_a, **__kw):
...
+class DefaultProgramValue(TimeStampedModel):
+    """
+    The model to store the default value set for all programs.
+    The image set stored in this model will be the default image presented through the API
+    when program specific image are not specified. This model is custom constructed to
+    only take one 1 instance of the model
+    """
+    banner_image = ResizingImageField(
+        path_template='program/banner/default',
+        sizes=[(1440, 480), (726, 242), (435, 145), (348, 116)],
```
rlucioni: I'd like to guarantee that these sizes always match those used for the Program model's banner_image field. What about assigning this list to a constant (e.g., `SIZES`) and using it in both places?

programs/apps/programs/admin.py
```diff
@@ -111,9 +113,25 @@ class ProgramCourseRunModeAdmin(admin.ModelAdmin):
...
 
 
+class DefaultProgramValueForm(forms.ModelForm):
+    """Admin form to handle the validation error for the model form"""
+
+    def clean(self):
+        if self.instance.__class__.objects.exists() and \
+                self.instance.__class__.objects.all()[0].id != self.instance.id:
+            raise forms.ValidationError(
+                'There is an instance of the DefaultProgramConfig model already exists')
```
rlucioni: "An instance of DefaultProgramConfig already exists."

programs/apps/programs/admin.py
```diff
@@ -111,9 +113,25 @@ class ProgramCourseRunModeAdmin(admin.ModelAdmin):
     readonly_fields = ('run_key',)
 
 
+class DefaultProgramValueForm(forms.ModelForm):
+    """Admin form to handle the validation error for the model form"""
+
+    def clean(self):
+        if self.instance.__class__.objects.exists() and \
+                self.instance.__class__.objects.all()[0].id != self.instance.id:
```
rlucioni: I think you can do this more concisely as:
```python
if self.instance.__class__.objects.exists():
    ...
```

programs/apps/api/serializers.py
```diff
@@ -269,11 +269,25 @@ class Meta(object):  # pylint: disable=missing-docstring
...
+        if len(default_value_instances) > 0:
+            default_url_items = default_value_instances[0].banner_image.resized_urls.items()
+        return default_url_items
+
     def get_banner_image_urls(self, instance):
         """
         Render public-facing URLs for the available banner images.
         """
         url_items = instance.banner_image.resized_urls.items()
+        if len(url_items) == 0:
```
rlucioni: More idiomatic would be:
```python
if not url_items:
    url_items = self._get_default_banner_images()
```

programs/apps/api/serializers.py
```diff
@@ -269,11 +269,25 @@ class Meta(object):  # pylint: disable=missing-docstring
     organizations = ProgramOrganizationSerializer(many=True, source='programorganization_set')
     course_codes = ProgramCourseCodeSerializer(many=True, source='programcoursecode_set', required=False)
 
+    def _get_default_banner_images(self):
+        """
+        Get the default banner image urls for all different sizes if available.
+        May return an empty list
+        """
+        default_url_items = []
+        default_value_instances = models.DefaultProgramValue.objects.all()
```
rlucioni: Since the system restricts `DefaultProgramValue` to a single instance, I think it would be cleaner to use `get()` here, handling possible exceptions as necessary: https://docs.djangoproject.com/en/1.9/ref/models/querysets/#get.

programs/apps/programs/models.py
```diff
@@ -275,3 +275,28 @@ def save(self, *_a, **__kw):
             raise ValidationError(_("Invalid course key."))
 
         return super(ProgramCourseRunMode, self).save()
+
+
+class DefaultProgramValue(TimeStampedModel):
```
rlucioni: I know you share my interest in clean names. What do you think of naming the model `ProgramDefault`? I think it's more readable. For example, `ProgramDefault.banner_image` vs `DefaultProgramValue.banner_image`.

programs/apps/programs/tests/test_models.py
```diff
@@ -247,3 +248,37 @@ def test_invalid_course_key(self, course_key):
             'Invalid course key.',
             context.exception.message
         )
+
+
+class TestDefaultProgramValue(TestCase):
+    """
+    Test case to validate the DefaultProgramBannerImage model
```
rlucioni: There is no model with this name.

programs/apps/programs/tests/test_models.py
```diff
@@ -247,3 +248,37 @@ def test_invalid_course_key(self, course_key):
...
+        """ Verify validation error if we attempt to create more than 1 default program banner image """
+        self._create_model_instance()
+        with self.assertRaises(ValidationError):
+            factories.DefaultProgramValueFactory.create()
+
+    def test_recreate(self):
+        """ Verify delete the only instance in db and new instance can be created successfully """
+        default_image = self._create_model_instance()
+        with self.assertRaises(ValidationError):
+            factories.DefaultProgramValueFactory.create()
```
rlucioni: Since you're already verifying that this `ValidationError` is raised in `test_create_another`, I think you can remove it here. The focus of this test case is verifying that the model can be recreated after deletion.

programs/apps/programs/tests/test_models.py
```diff
@@ -247,3 +248,37 @@ def test_invalid_course_key(self, course_key):
...
+        default_image.delete()
+        default_image = self._create_model_instance()
```
rlucioni: No need to assign the return value to a variable here.

programs/apps/programs/tests/test_models.py
```diff
@@ -247,3 +248,37 @@ def test_invalid_course_key(self, course_key):
...
+    """
+
+    def _create_model_instance(self):
+        """
+        Helper to create the default program banner image model instance
+        """
+        default_value_instance = factories.DefaultProgramValueFactory.create()
+        default_value_instance.banner_image = make_banner_image_file('test.jpg')
+        default_value_instance.save()
+        self.assertGreater(len(default_value_instance.banner_image.field.sizes), 0)
```
rlucioni: Even better, check that the number of sizes matches the length of the `SIZES` constant I referred to earlier.

programs/apps/api/v1/tests/test_views.py
```diff
@@ -350,6 +340,39 @@ def assert_correct_banner_image_urls(self, url_prepend=''):
...
         self.assertEqual(response.data[u'banner_image_urls'], expected_urls)
 
+    def assert_correct_default_banner_image_urls(self, url_prepend='', set_program_banner=False):
+        """
+        DRY test helper.  Ensure that the serializer generates a complete set
+        of absolute URLs for the default banner_image when the banner image on program is not set.
+        If the flag set_program_banner is defined, then make sure the banner image for program is set
+        and returned
+        """
+        default_program_value = DefaultProgramValueFactory.create()
```
rlucioni: You can use the Factory class as a shortcut for the default build strategy:
```python
default_program_value = DefaultProgramValueFactory()
```
http://factoryboy.readthedocs.io/en/latest/#using-factories

programs/apps/api/v1/tests/test_views.py
```diff
@@ -350,6 +340,39 @@ def assert_correct_banner_image_urls(self, url_prepend=''):
...
+        default_program_value.banner_image = make_banner_image_file('default_test_filename.jpg')
+        default_program_value.save()
+
+        program = ProgramFactory.create(status=ProgramStatus.ACTIVE)
```
rlucioni:
```python
program = ProgramFactory(status=ProgramStatus.ACTIVE)
```

programs/apps/api/v1/tests/test_views.py
```diff
@@ -350,6 +340,39 @@ def assert_correct_banner_image_urls(self, url_prepend=''):
...
+        if set_program_banner:
+            program.banner_image = make_banner_image_file('test_filename.jpg')
+        program.save()
```
rlucioni: This `save` is only necessary if a change is made in the preceding if block. I'd indent it.

programs/apps/api/v1/tests/test_views.py
```diff
@@ -350,6 +340,39 @@ def assert_correct_banner_image_urls(self, url_prepend=''):
...
+
+        expected_urls = {
+            'w{}h{}'.format(*size): '{}{}__{}x{}.jpg'.format(
+                url_prepend,
+                banner_image_instance.url,
+                *size)
+            for size in banner_image_instance.field.sizes
+        }
+        self.assertEqual(response.data[u'banner_image_urls'], expected_urls)
+        default_program_value.delete()
```
rlucioni: I don't think this is necessary, the database should be cleaned up during teardown.

programs/apps/programs/models.py
bderusha: @schenedx here is an interesting implementation as well

https://github.com/lazybird/django-solo/blob/master/solo/models.py#L12-L23

programs/apps/programs/tests/test_models.py
```diff
@@ -247,3 +248,37 @@ def test_invalid_course_key(self, course_key):
...
+
+
+class TestDefaultProgramValue(TestCase):
+    """
+    Test case to validate the DefaultProgramBannerImage model
+    """
+
+    def _create_model_instance(self):
+        """
+        Helper to create the default program banner image model instance
```
bderusha: Same comment as @rlucioni above. Update text to reflect new model name

bderusha:
@rlucioni was pretty thorough here.  I added a few comments of my own, but I agree with all of his.  👍 Once they are addressed.

programs/apps/programs/models.py
rlucioni: @bderusha nice find! @schenedx consider subclassing [django-solo](https://github.com/lazybird/django-solo)'s `SingletonModel`. It'll get us some nice caching and an admin interface that's aware you only have one object.

programs/apps/api/serializers.py
schenedx: @rlucioni I have considered using `get` here. The reason why I didn't use that is because there is no good unique identifier to use the `get` with. The only identifier I see is the pk. However, pk would change when user deleted the only value and recreate. With each recreate, the pk would increment by 1.

programs/apps/api/serializers.py
schenedx: @rlucioni Fair point, will update

programs/apps/api/v1/tests/test_views.py
schenedx: @rlucioni Understand the shorthand. I do like this way better tho. It's clear what I am doing, so easier to read and understand.

programs/apps/api/v1/tests/test_views.py
schenedx: @rlucioni Good point. Will update

programs/apps/api/v1/tests/test_views.py
schenedx: Same point as above. I like the create keyword to make code easier to read.

programs/apps/programs/models.py
schenedx: @rlucioni Ah, better name. I agree. Will try to update to it.

programs/apps/programs/models.py
schenedx: @rlucioni Good point. Will try. But I suspect that will generate some migration may be intrusive to the existing ResizableImageField?

programs/apps/programs/tests/test_models.py
schenedx: @rlucioni That was the original name of the model I changed from. Bad one huh. Will update

programs/apps/programs/tests/test_models.py
schenedx: @rlucioni Will update

programs/apps/programs/tests/test_models.py
schenedx: @rlucioni Will update

programs/apps/api/serializers.py
rlucioni: @schenedx you don't need to use any identifiers with `get()` if you're dealing with singleton models. Just call it and you'll get back the only object in the system, if it exists.

programs/apps/programs/models.py
rlucioni: @schenedx it shouldn't, since the value being used to define the sizes remains unchanged (i.e., it's still the same list of tuples). However, trying it is the best way to find out.

programs/apps/api/serializers.py
```diff
@@ -269,11 +269,25 @@ class Meta(object):  # pylint: disable=missing-docstring
     organizations = ProgramOrganizationSerializer(many=True, source='programorganization_set')
     course_codes = ProgramCourseCodeSerializer(many=True, source='programcoursecode_set', required=False)
 
+    def _get_default_banner_images(self):
+        """
+        Get the default banner image urls for all different sizes if available.
+        May return an empty list
```
rlucioni: Nitpick: Use Google-style in docstrings like these:
```python
def _get_default_banner_images(self):
    """Get default banner image URLs.

    Returns:
        list of tuples if default banner image has been configured. Empty list otherwise.
    """
```

programs/apps/programs/models.py
```diff
@@ -275,3 +279,18 @@ def save(self, *_a, **__kw):
             raise ValidationError(_("Invalid course key."))
 
         return super(ProgramCourseRunMode, self).save()
+
+
+class ProgramDefault(SingletonModel):
+    """
+    The model to store the default value set for all programs.
+    The image set stored in this model will be the default image presented through the API
+    when program specific image are not specified. Only one instance of this model would exist.
```
rlucioni: Consider cleaning this up a bit:
> Model used to store default program configuration. This includes a default, resizable banner image which is used if a given program has not specified its own banner image. This model is a singleton - only one instance exists.

requirements/base.txt
```diff
@@ -20,3 +20,4 @@ Markdown==2.6.2
 piexif==1.0.3
 Pillow==3.1.1
 pytz==2015.4
+django-solo==1.1.2
```
rlucioni: Nitpick: please keep requirements in alphabetical order.

programs/apps/programs/admin.py
```diff
@@ -1,8 +1,10 @@
 """Admin for programs models."""
 from django import forms
 from django.contrib import admin
+
```
rlucioni: Nitpick: This newline isn't necessary. All third-party imports should be grouped together.

programs/apps/programs/models.py
```diff
@@ -74,7 +78,7 @@ class Program(TimeStampedModel):
 
     banner_image = ResizingImageField(
         path_template='program/banner/{uuid}',
-        sizes=[(1440, 480), (726, 242), (435, 145), (348, 116)],
+        sizes=RESIZABLE_IMAGE_SIZES,
```
rlucioni: @schenedx were you able to verify that this won't require a migration?

rlucioni:
@schenedx looks nice. Just a few minor comments. 👍 otherwise.

programs/apps/programs/models.py
schenedx: Yes, this won't require a migration

PR #133, opened 2016-05-09
ECOM-4403 fix the bug with clean stale images

schenedx:
The line in PR https://github.com/edx/programs/pull/128/files#diff-b4b27d9e5b70b25e71fa7e0fd5762a21R118 is problematic if we use S3BotoStorageFile implementation of the Django Storage API. That implementation do not have a "created_time". Instead, it has the "modified_time" property for this purpose.

This fixes that

@rlucioni Please review


rlucioni:
:+1:

PR #134, opened 2016-05-10
ECOM-4309 update the Organization model to remove the display_name unique constraint

schenedx:
@rlucioni Please review
@AlasdairSwan FYI

Tested locally:
 - Made sure with a dupe long name in the LMS organization data, the sync command fails before the change
 - After the change, make sure the sync command succeeds

### DIFF
--- a/programs/apps/programs/tests/test_sync_orgs.py
+++ b/programs/apps/programs/tests/test_sync_orgs.py
@@ -24,7 +24,7 @@ class SyncOrgsTests(TestCase):
     DISPLAY_NAME = 'Organization {} Display Name'
     KEY = 'Organization{}X'
 
-    def _get_body(self, page):
+    def _get_body(self, page, name_override=None):
         next_page = page + 1 if page < self.PAGE_COUNT else None
         previous_page = page - 1 if page > 1 else None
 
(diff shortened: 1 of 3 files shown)


programs/apps/programs/tests/test_sync_orgs.py
```diff
@@ -24,7 +24,7 @@ class SyncOrgsTests(TestCase):
     DISPLAY_NAME = 'Organization {} Display Name'
     KEY = 'Organization{}X'
 
-    def _get_body(self, page):
+    def _get_body(self, page, name_override=''):
```
rlucioni: Nitpick: it's more idiomatic to use `None` as a default kwarg.

programs/apps/programs/tests/test_sync_orgs.py
```diff
@@ -53,14 +53,16 @@ def _get_body(self, page):
 
         return json.dumps(body)
 
-    def _mock_organizations_api(self):
+    def _mock_organizations_api(self, org_name_override=''):
```
rlucioni: Nitpick: it's more idiomatic to use `None` as a default kwarg.

rlucioni:
Just a few minor notes, 👍 otherwise.

PR #135, opened 2016-05-31
Remove django-waffle (unused)

rlucioni:
I was going to upgrade to 0.11.1 but then realized this project doesn't actually use Waffle.

@edx/ecommerce


clintonb:
👍 

Note that the tables will remain in the database, which isn't necessarily a bad thing.

rlucioni:
@clintonb do you know if that will cause problems should we chose to re-introduce the dependency at a later date?

clintonb:
It shouldn't. Those migrations will be skipped.

schenedx:
👍

PR #136, opened 2016-06-09
Add deprecation warning to completion endpoint

rlucioni:
No longer necessary after work done in https://github.com/edx/edx-platform/pull/12700.

@schenedx @mikedikan FYI.


schenedx:
👍

PR #137, opened 2016-06-09
Delete program completion utility and endpoint

rlucioni:
This functionality has been moved to the LMS and is no longer necessary within programs. ECOM-4490.

@schenedx @mikedikan FYI. I won't merge this until changes from https://github.com/edx/edx-platform/pull/12700 are deployed.


schenedx:
@rlucioni If you are deleting these endpoints, why merge in https://github.com/edx/programs/pull/136?

rlucioni:
#136 deprecates the endpoint. This PR removes it. You're right that #136 wasn't totally necessary given this PR, but I wanted a message in the code prior to this PR landing. If this were a public-facing API and we weren't the only consumers of it, the most correct thing to do would be to bump the version of the API, deprecate v1, and eventually delete it.

schenedx:
@rlucioni Glad to see you are approaching this thoroughly.
Here are somethings I think made https://github.com/edx/programs/pull/136 unnecessary:
1. Programs IDA is not really supported in the openedx community. 
2. With your deprecation PR, you did not deploy the changes to production. This means the deprecation is only going to be visible in the commit history of this repo. That has very small, if not no value.
3. We are looking to replace this IDA with Course Catalog IDA. I think the less effort we put into this IDA, the better at the moment.

rlucioni:
@schenedx all fair points. My thinking at the time was that having a warning in the code would prevent confusing others in case I wasn't around to explain, before the endpoint gets deleted. In retrospect, it may have made more sense to just open this PR; deleting code on this end will make any move that much easier.

rlucioni:
There have been three successful edx-platform releases since this functionality stopped being used. At this point, it's safe to remove this code without fear of a rollback requiring that it be available. I'll merge once the build passes.

PR #138, opened 2016-06-13
Make user creation during JWT authentication more robust

rlucioni:
Even with MySQL set to use the READ COMMITTED isolation level, get_or_create during JWT authentication sometimes raises an IntegrityError and the user in question fails to appear in a subsequent get() call. These changes address this issue by disabling atomic requests for the programs API and introducing more robust retry logic meant to prevent IntegrityErrors from causing authenticated requests to fail. ECOM-4506.

@cpennington please review. @mikedikan @schenedx FYI.

### DIFF
--- a/programs/apps/api/authentication.py
+++ b/programs/apps/api/authentication.py
@@ -1,8 +1,8 @@
 """
 Authentication logic for REST API.
 """
-
 import logging
+import time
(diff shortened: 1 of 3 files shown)


cpennington:
:+1:

PR #140, opened 2016-06-22
WIP: Experiments with the release pipeline.

doctoryes:
Creating this PR against the branch so that tests will run with each commit.

@edx/pipeline-team FYI


clintonb:
We now have release pipelines!

PR #145, opened 2016-07-01
Fix to protocol for OIDC redirect in setup docs

mikedikan:


mikedikan:
@rlucioni as discussed

rlucioni:
:shipit:

PR #146, opened 2016-07-05
Exposing program UUID field

clintonb:
ECOM-4884


clintonb:
@bderusha please review

PR #147, opened 2016-07-05
Remove edx-icons usage in Programs

alisan617:
Refs [TNL-4814](https://openedx.atlassian.net/browse/TNL-4814)

[Sandbox](https://alisan617.sandbox.edx.org)

## Description
- Remove edx-icons usage in Programs.
- Update edX Pattern Library version in bower package.
- Remove icon-fallback wrapper span.
- Update templates to use new class.

## Post-review
- [ ] Squash commits into discrete sets of changes with descriptive commit messages.

## Reviewers
- [x] @AlasdairSwan  
- [ ] @cahrens  

If you've been tagged for review, please check your corresponding box once you've given the :+1:.

We'll also automatically suggest reviewers for this PR based on the code it touches.


AlasdairSwan:
👍

programs/static/templates/course_details.underscore
```diff
@@ -3,21 +3,19 @@
...
-        <span class="icon-fallback icon-fallback-text">
-            <span class="icon icon-trash-o" aria-hidden="true"></span>
-            <span class="text"><%= interpolate(
-                gettext('Remove %(name)s from the program'),
-                { name: display_name },
-                true
-            ) %></span>
-        </span>
+        <span class="icon fa fa-trash-o" aria-hidden="true"></span>
+        <span class="sr-only"><%= interpolate(
```
cahrens: Note that these files should be using <%- instead of <%= to properly handle escaping. I wouldn't recommend fixing this as part of this PR, but it would be good to create a JIRA ticket for XSS updates to the Programs templates.

cahrens:
This looks reasonable, but I do think it would be good to have someone who is familiar with the feature do a little manual testing once the sandbox is configured. I don't fully understand all the pattern library/CSS changes, but I'm assuming @AlasdairSwan does understand them.

👍 with the above caveats

programs/static/templates/course_details.underscore
alisan617: Good call. Jira ticket has been created https://openedx.atlassian.net/browse/TNL-4938

alisan617:
cc @schenedx

schenedx:
👍

PR #149, opened 2016-07-05
Update edx-drf-extensions

bderusha:
Update drf-extensions to use multiple issuers

@clintonb @mattdrayer please review

### DIFF
--- a/programs/settings/production.py
+++ b/programs/settings/production.py
@@ -46,7 +46,16 @@ for override, value in DB_OVERRIDES.iteritems():
     DATABASES['default'][override] = value
 
 JWT_AUTH.update({
-    'JWT_SECRET_KEY': SOCIAL_AUTH_EDX_OIDC_SECRET,
-    'JWT_ISSUER': SOCIAL_AUTH_EDX_OIDC_URL_ROOT,
-    'JWT_AUDIENCE': SOCIAL_AUTH_EDX_OIDC_KEY,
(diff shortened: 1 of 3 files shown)


programs/settings/production.py
```diff
@@ -46,7 +46,16 @@
     DATABASES['default'][override] = value
 
 JWT_AUTH.update({
-    'JWT_SECRET_KEY': SOCIAL_AUTH_EDX_OIDC_SECRET,
-    'JWT_ISSUER': SOCIAL_AUTH_EDX_OIDC_URL_ROOT,
-    'JWT_AUDIENCE': SOCIAL_AUTH_EDX_OIDC_KEY,
+    'JWT_ISSUERS': [
```
clintonb: This entire list should come from configuration, rather than being pieced together here.

requirements/base.txt
```diff
@@ -10,7 +10,7 @@ djangorestframework==3.2.3
 djangorestframework-jwt==1.8.0
 django-rest-swagger==0.3.4
 edx-auth-backends==0.2.3
-edx-drf-extensions==0.4.1
+edx-drf-extensions==1.1.0
```
mattdrayer: I think it's 1.1.1 now --> https://github.com/edx/edx-drf-extensions/releases/tag/1.1.1

clintonb:
Replaced by https://github.com/edx/programs/pull/150.

PR #152, opened 2016-07-14
Unpack media storage settings when applying config overrides

rlucioni:
@e0d @maxrothman @clintonb 

This was removed in https://github.com/edx/programs/commit/ec2a6cd8a01bff561e8d0b61988be68c8ea1d949#diff-2ca989f014687f3223a751f4dbc2756eL34.

### DIFF
--- a/programs/settings/production.py
+++ b/programs/settings/production.py
@@ -23,7 +23,7 @@ LOGGING = get_logger_config()
 
 # Keep track of the names of settings that represent dicts. Instead of overriding the values in base.py,
 # the values read from disk should UPDATE the pre-configured dicts.
-DICT_UPDATE_KEYS = ('JWT_AUTH', 'MEDIA_STORAGE_BACKEND',)
+DICT_UPDATE_KEYS = ('JWT_AUTH',)
 
 # This may be overridden by the YAML in PROGRAMS_CFG, but it should be here as a default.
 MEDIA_STORAGE_BACKEND = {}
@@ -41,6 +41,9 @@ with open(CONFIG_FILE) as f:
             vars()[key].update(value)
 
     vars().update(config_from_yaml)
+
+    # Unpack media storage settings.
+    vars().update(MEDIA_STORAGE_BACKEND)
 
 DB_OVERRIDES = dict(
     PASSWORD=environ.get('DB_MIGRATION_PASS', DATABASES['default']['PASSWORD']),


clintonb:
Why/how is `MEDIA_STORAGE_BACKEND` different from `JWT_AUTH`? The idea there is that dictionaries are updated from settings rather than replaced.

clintonb:
Also, is there an acceptance test we can update to confirm this is fixed?

rlucioni:
@clintonb the contents of `MEDIA_STORAGE_BACKEND` need to be unpacked in order to override other settings. See [here](https://github.com/edx/configuration/blob/master/playbooks/roles/programs/defaults/main.yml#L94) for an example of how that dictionary might look. Note how `MEDIA_URL` and `MEDIA_ROOT` are contained in that dictionary.

rlucioni:
@clintonb I'm not aware of existing tests covering the loading/overriding of settings in production environments. If we had such tests, this would likely have been caught. I'd be happy to look into adding them, though I'd prefer to take care of that separately.

clintonb:
Wait...why are we using an unorthodox setting like `MEDIA_STORAGE_BACKEND`? There were no comments on that setting, so I assumed it was another `dict` to be set, similar to `JWT_AUTH`. Can we use the actual settings—`DEFAULT_FILE_STORAGE`, `MEDIA_ROOT`, `MEDIA_URL`—and avoid this confusion in the future?

rlucioni:
@clintonb I remember it being a deliberate choice, though I'm having trouble remembering/finding a record of the justification for it. I remember @maxrothman was pretty opinionated about this approach. @maxrothman any chance you remember the reasoning around this?

maxrothman:
We'd prefer to group all the settings necessary to switch between backends in configuration so they can be swapped easily. `MEDIA_STORAGE_BACKEND` contains all the standard settings, it just wraps them so it's easy to tell which ones go together.

clintonb:
How often do we switch backends? This is a solution to a non-existent problem, and it's too clever for its own good, hence our having this conversation now.

e0d:
I'm :+1: to release this and debate the best factoring later.

clintonb:
👍 (pending):

1. Remove `MEDIA_STORAGE_BACKEND ` from `DICT_UPDATE_KEYS`
2. Create a ticket to replace `MEDIA_STORAGE_BACKEND` with the individual settings.

maxrothman:
👍 for this change

@clintonb we change backends constantly: we use local file storage for devstacks/sandboxes and s3 for prod/stage/loadtest. Members of the community use openstack, local file storage, azure, and lots of other solutions.

clintonb:
@maxrothman the same case could be made for practically every setting in edx/configuration. There isn't a good case for treating these settings any differently than others.

maxrothman:
@clintonb I've already explained why this is a special case. Most other settings don't come in big groups and aren't changed in a decent chunk of deployments. I understand that's it's ugly to have a special case here, but configuration is the interface most open-source users use to customize their deployment. That interface should make it easy to do the most common things people want to do, one of which is switch storage backends. Switching storage backends is otherwise a complex and fragile process.

PR #153, opened 2016-07-19
Revision to program data model to allow for a course to be part of mu…

mikedikan:
…ltiple programs and adding the micromaster type of program

Please see https://openedx.atlassian.net/browse/ECOM-4905 for additional details.

### DIFF
--- a/programs/apps/programs/constants.py
+++ b/programs/apps/programs/constants.py
@@ -8,6 +8,7 @@ class ProgramCategory(object):
     """Allowed values for Program.category"""
 
     XSERIES = "xseries"
+    MICROMASTERS = "micromasters"
 
 
 class ProgramStatus(object):
--- a/programs/apps/programs/tests/test_models.py
+++ b/programs/apps/programs/tests/test_models.py
@@ -105,24 +105,34 @@ class TestProgramCourseCode(TestCase):
         factories.ProgramOrganizationFactory.create(program=self.program, organization=self.org)
         super(TestProgramCourseCode, self).setUp()
 
-    def test_one_program_max(self):
+    def test_multi_program_course_association(self):
         """
-        Ensure that a CourseCode cannot be associated with more than one program
-        (the relationship is modeled as m2m, but initially, we will only allow
-        one organization to be associated).
+        Ensure that a CourseCode can be associated with more than one program
+        (the relationship is modeled as m2m).
         """
-        course_code = factories.CourseCodeFactory.create(organization=self.org)
-        orig_pgm_course = factories.ProgramCourseCodeFactory.create(program=self.program, course_code=course_code)
 
+        # Create an initial course code and attach it to a program
+        course_code = factories.CourseCodeFactory.create(organization=self.org)
+        factories.ProgramCourseCodeFactory.create(program=self.program, course_code=course_code)
+
+        # Ensure that the course code can not be used for different organizations...
         program2 = factories.ProgramFactory.create()
+        organization2 = factories.OrganizationFactory.create()
+        factories.ProgramOrganizationFactory.create(
+            program=program2, organization=organization2
+        )
         pgm_course = factories.ProgramCourseCodeFactory.build(program=program2, course_code=course_code)
         with self.assertRaises(ValidationError) as context:
             pgm_course.save()
-        self.assertEqual('Cannot associate multiple programs with a course code.', context.exception.message)
+        self.assertEqual(
+            'Course code must be offered by the same organization offering the program.',
+            context.exception.message
+        )
 
-        # make sure it works to reassign the existing association to the other program
-        orig_pgm_course.program = program2
-        orig_pgm_course.save()
+        # But can be used for multiple programs offered by the same organization
+        program3 = factories.ProgramFactory.create()
+        factories.ProgramOrganizationFactory.create(program=program3, organization=self.org)
+        factories.ProgramCourseCodeFactory.create(program=program3, course_code=course_code)
 
     def test_position(self):
         """
--- a/docs/prerequisites.rst
+++ b/docs/prerequisites.rst
@@ -2,8 +2,11 @@ Software Prerequisites
 ======================
 
 - Clone of `edx/programs <https://github.com/edx/programs>`_
+
 - Python 2.7.x (and pip)
+
 - `Node.js 4.4.x`_ (and npm) (download or use nvm - explained below)
+
 .. _Node.js 4.4.x: https://nodejs.org/download/release/v4.4.7/
 
 
--- a/docs/testing.rst
+++ b/docs/testing.rst
@@ -1,6 +1,20 @@
 Testing
 =======
 
+To run the unit tests and get a code coverage report:
+
+.. code-block:: bash
+
+    $ make test
+
+To run a single test, remove the compiled python files picked up by the test launcher and provide the path to the
+single test:
+
+.. code-block:: bash
+
+    $ make clean
+    $ ./manage.py test programs/path/to/your/python_test.py --settings=programs.settings.test
+
 The command below runs the Python tests and code quality validation—Pylint and PEP8.
 
 .. code-block:: bash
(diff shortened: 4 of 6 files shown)


mikedikan:
@schenedx or @rlucioni please review

programs/apps/programs/constants.py
```diff
@@ -8,6 +8,7 @@ class ProgramCategory(object):
     """Allowed values for Program.category"""
 
     XSERIES = "xseries"
+    MICROMASTERS = "micromasters"
```
rlucioni: This is a great opportunity to clean up some debt. Take a look at my comment at https://github.com/edx/course-discovery/pull/143#discussion_r69153015.

schenedx: The new library looks good. 
But since we are looking to retire programs IDA in the near future, I don't think we should spend more energy here to make the code look good.
I think what @mikedikan have here is good enough.

schenedx:
👍

programs/apps/programs/tests/test_models.py
```diff
@@ -105,25 +105,6 @@ def setUp(self):
         factories.ProgramOrganizationFactory.create(program=self.program, organization=self.org)
         super(TestProgramCourseCode, self).setUp()
 
-    def test_one_program_max(self):
```
rlucioni: Instead of removing this, I'd leave it in place and modify it so that it verifies that a course code *can* be associated with more than one program.

programs/apps/programs/constants.py
rlucioni: That's fair, moving the models will be another good opportunity to make this change.

rlucioni:
@mikedikan I think you're going to need to create a migration to account for your model changes. https://docs.djangoproject.com/en/1.9/topics/migrations/#workflow

mikedikan:
@schenedx @rlucioni addressed review comments

programs/apps/programs/tests/test_models.py
```diff
@@ -105,24 +105,32 @@ def setUp(self):
...
-        (the relationship is modeled as m2m, but initially, we will only allow
-        one organization to be associated).
+        Ensure that a CourseCode can be associated with more than one program
+        (the relationship is modeled as m2m).
         """
-        course_code = factories.CourseCodeFactory.create(organization=self.org)
-        orig_pgm_course = factories.ProgramCourseCodeFactory.create(program=self.program, course_code=course_code)
 
+
+        #Ensure that a CourseCode can not be used for different organizations...
```
rlucioni: pep8 is failing your build because of these comments. There should be a space between the `#` and the comment. See https://www.python.org/dev/peps/pep-0008/#inline-comments for more, and run `make quality` locally to verify you fixed everything.

programs/apps/programs/tests/test_models.py
```diff
@@ -105,24 +105,32 @@ def setUp(self):
...
+        course_code = factories.CourseCodeFactory.create(organization=self.org)
         program2 = factories.ProgramFactory.create()
+        organization2 = factories.OrganizationFactory.create()
+        program_organization2 = factories.ProgramOrganizationFactory.create(program=program2, organization=organization2)
+        program2.save()
+        organization2.save()
+        program_organization2.save()
```
rlucioni: You don't need these explicit saves. `create` saves automatically for you.

programs/apps/programs/tests/test_models.py
```diff
@@ -105,24 +105,32 @@ def setUp(self):
...
+        self.assertEqual('Course code must be offered by the same organization offering the program.', context.exception.message)
 
-        # make sure it works to reassign the existing association to the other program
-        orig_pgm_course.program = program2
-        orig_pgm_course.save()
+        #But can be used for multiple programs offered by the same organization
+        program3 = factories.ProgramFactory.create()
+        program_organization3 = factories.ProgramOrganizationFactory.create(program=program3, organization=self.org)
+        program3.save()
+        program_organization3.save()
```
rlucioni: Same here, these saves can be removed.

mikedikan:
@rlucioni addressed latest - please confirm

programs/apps/programs/tests/test_models.py
```diff
@@ -105,24 +105,27 @@ def setUp(self):
...
             pgm_course.save()
-        self.assertEqual('Cannot associate multiple programs with a course code.', context.exception.message)
+        self.assertEqual('Course code must be offered by the same organization offering the program.', context.exception.message)
 
-        # make sure it works to reassign the existing association to the other program
-        orig_pgm_course.program = program2
-        orig_pgm_course.save()
+        #But can be used for multiple programs offered by the same organization
+        program3 = factories.ProgramFactory.create()
+        program_organization3 = factories.ProgramOrganizationFactory.create(program=program3, organization=self.org)
```
schenedx: I don't see the call to `ProgramCourseCodeFactory` for Program 3 here. Is that intentional? How does program 3 associated itself with the same course code?

mikedikan: Yes, you're right.  I've fixed and recommitted.

mikedikan:
@schenedx @rlucioni addressed comment issue with spacing that I previously missed.

programs/apps/programs/tests/test_models.py
```diff
@@ -105,24 +105,29 @@ def setUp(self):
...
+        self.assertEqual('Course code must be offered by the same organization offering the program.', context.exception.message)
 
-        # make sure it works to reassign the existing association to the other program
-        orig_pgm_course.program = program2
-        orig_pgm_course.save()
+        # But can be used for multiple programs offered by the same organization
+        program3 = factories.ProgramFactory.create()
+        program_organization3 = factories.ProgramOrganizationFactory.create(program=program3, organization=self.org)
+        pgm_course3 = factories.ProgramCourseCodeFactory.build(program=program3, course_code=course_code)
+        pgm_course3.save()
```
rlucioni: You should be able to use `create` instead of `build` and `save` here.

programs/apps/programs/tests/test_models.py
```diff
@@ -105,24 +105,29 @@ def setUp(self):
...
-        Ensure that a CourseCode cannot be associated with more than one program
-        (the relationship is modeled as m2m, but initially, we will only allow
-        one organization to be associated).
+        Ensure that a CourseCode can be associated with more than one program
+        (the relationship is modeled as m2m).
         """
-        course_code = factories.CourseCodeFactory.create(organization=self.org)
-        orig_pgm_course = factories.ProgramCourseCodeFactory.create(program=self.program, course_code=course_code)
 
+
```
rlucioni: This extra newline is causing your build failure. Again, you can run `make quality` locally to verify your syntax is clean.

rlucioni:
👍 once build is passing.

mikedikan:
@schenedx @rlucioni make quality now passes - please give me a final thumb so i can merge master and rebase 🙏

programs/apps/programs/tests/test_models.py
```diff
@@ -105,24 +105,32 @@ def setUp(self):
...
+            context.exception.message
+        )
 
-        # make sure it works to reassign the existing association to the other program
-        orig_pgm_course.program = program2
-        orig_pgm_course.save()
+        # But can be used for multiple programs offered by the same organization
+        program3 = factories.ProgramFactory.create()
+        factories.ProgramOrganizationFactory.create(program=program3, organization=self.org)
+        factories.ProgramCourseCodeFactory.create(program=program3, course_code=course_code)
```
rlucioni: I don't think this is enough to test that a course code can be associated with more than one program. This is the only successful association of a course code with a program in this test case. You need to do it one more time with another program (e.g., `self.program`).

mikedikan:
@rlucioni @schenedx any last words before i merge - can i get some thumbs?

rlucioni:
👍

schenedx:
👍

rlucioni:
🎉

PR #155, opened 2016-07-20
ECOM-4904 remove the program editor from Programs IDA

schenedx:
@rlucioni @mikedikan Please review

This is a removal of all frontend assets from the Program IDA


.travis.yml
```diff
@@ -5,18 +5,13 @@ sudo: false
...
 before_install:
   - "export DISPLAY=:99.0"
   - "sh -e /etc/init.d/xvfb start"
 install:
   - pip install -U pip wheel codecov
   - pip install -r requirements/test.txt
-  - make requirements.js
 script:
-    # Compile assets and run validation
   - make static -e DJANGO_SETTINGS_MODULE="programs.settings.test"
```
rlucioni: You can remove this.

Makefile
```diff
@@ -22,52 +21,38 @@ help:
...
-
-requirements: requirements.js
+requirements:
 	pip install -qr requirements/local.txt --exists-action w
 
 test: clean
 	REUSE_DB=1 coverage run ./manage.py test programs --settings=programs.settings.test
 	coverage report
 
+static:
```
rlucioni: I think this entire target can go. There are no more static assets to collect.

rlucioni:
@schenedx looking good. Could you please also remove CORS-related dependencies and config? Since the authoring app now lives in Studio, this support is no longer necessary. Take a look at the PR where I introduced it to figure out what to remove: https://github.com/edx/programs/pull/51.

Makefile
```diff
@@ -22,52 +21,38 @@ help:
 	@echo "  requirements               install requirements for local development"
 	@echo "  requirements.js            install JavaScript requirements for local development"
```
mikedikan: Remove requirements.js from the help text

rlucioni:
👍 once comments addressed.

rlucioni:
These changes have also shaved more than a minute off of the project build time! 🎆

schenedx:
@rlucioni Well, we are putting more unit tests and sass into studio, which is already too big 😞

rlucioni:
@schenedx true, though to be fair Studio already installs a lot of the assets necessary for the app via npm and bower. I think those two took up most of the build time before.

rlucioni:
@schenedx speaking of which, you can also get rid of `.nvmrc` since this project no longer requires node.

rlucioni:
Before merging this, please update https://github.com/edx/configuration/blob/master/playbooks/roles/programs/tasks/main.yml so that CI builds and deployments continue to function.

schenedx:
@rlucioni Thank you for the reminder. The configuration repo changes are made at https://github.com/edx/configuration/pull/3262

rlucioni:
🎉

mikedikan:
👍

PR #156, opened 2016-07-22
Remove django-compressor

rlucioni:
We no longer use this tool. Counterpart to https://github.com/edx/configuration/pull/3267.

@schenedx @mikedikan please review.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,7 +30,7 @@ clean:
 	coverage erase
 
 requirements:
-	pip install -qr requirements/local.txt --exists-action w
+	pip install -r requirements/local.txt
 
 test: clean
 	REUSE_DB=1 coverage run ./manage.py test programs --settings=programs.settings.test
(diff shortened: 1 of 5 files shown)


schenedx:
@rlucioni Are you sure the existing django-rest-framework does not use compressor?
If so, 👍

rlucioni:
@schenedx yup, DRF doesn't require compressor.

Makefile
```diff
@@ -30,7 +30,7 @@ clean:
 	coverage erase
 
 requirements:
-	pip install -qr requirements/local.txt --exists-action w
```
mikedikan: Can you explain these changes to silence the output and the wipe action if a package exists?

rlucioni: I'm getting rid of both. Having log output is helpful: we want to see what's being installed (or not installed), especially when trying to debug something. Specifying an [exists-action](https://pip.pypa.io/en/stable/reference/pip/#exists-action-option) only appears to be relevant when installing packages using a [VCS system](https://pip.pypa.io/en/stable/reference/pip_install/#vcs-support) (e.g., git), something we're not doing in this project (we install from PyPI).

mikedikan:
👍

PR #158, opened 2016-07-25
Case category constants for display

rlucioni:
Ultimately, these will be replaced by a proper ProgramCategory model. In the meantime, correcting the casing here does away with the need to force consumers of this API to parse and hardcode references to these categories. Django doesn't enforce choices on a database level; the migration exists to appease Django 1.8's migrations framework. Part of ECOM-5018.

Since we intend to roll this into the catalog, I realize that we should generally avoid making changes to this code. However, I think the benefits of doing this make this case an exception. I'm going to create corresponding PRs for Studio and the LMS.

@mikedikan @schenedx please review. @bderusha @clintonb FYI.

### DIFF
--- a/programs/apps/programs/constants.py
+++ b/programs/apps/programs/constants.py
@@ -7,14 +7,14 @@ from __future__ import unicode_literals
 class ProgramCategory(object):
     """Allowed values for Program.category"""
 
-    XSERIES = "xseries"
-    MICROMASTERS = "micromasters"
+    XSERIES = 'XSeries'
+    MICROMASTERS = 'MicroMasters'
 
 
 class ProgramStatus(object):
     """Allowed values for Program.status"""
 
-    UNPUBLISHED = "unpublished"
-    ACTIVE = "active"
-    RETIRED = "retired"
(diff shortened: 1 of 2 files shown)


clintonb:
This will break any code that has come to expect a lower-cased category.

rlucioni:
@clintonb my plan is to 1) release updates to API consumers so that they lowercase whatever comes across, 2) update the programs data, and 3) release updates to consumers so they expect correctly-cased strings.

schenedx:
👍

mikedikan:
I'm confused - If we just deploy this change - won't it break the programs consumers?  Shouldn't we deploy the code update first on the consumer side so that it lowercases these values coming from the programs IDA?

rlucioni:
@mikedikan as long as we wait to update the category data actually stored in the programs db, it shouldn't matter one way or the other. I'd be happy to chat in person if you're still confused.

mikedikan:
Sure - i'll message you on hipchat

mikedikan:
👍  (after in person chat)

PR #160, opened 2016-08-02
Set the copyright year in docs dynamically

rlucioni:
@lamagnifica @andy-armstrong

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -12,6 +12,7 @@
 # serve to show the default.
 
 import os
+import datetime
 
 # on_rtd is whether we are on readthedocs.org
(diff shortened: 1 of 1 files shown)


docs/conf.py
```diff
@@ -48,8 +49,8 @@
 master_doc = 'index'
 
 # General information about the project.
-project = u'Programs'
-copyright = u'2015, edX'
+project = 'Programs'
+copyright = '{year}, edX'.format(year=datetime.datetime.now().year)
```
lamagnifica: fancy! I didn't know you could do that!

rlucioni: 😄

PR #162, opened 2016-08-22
Add an OEP-2 compliant openedx.yaml file

cpennington:
This adds an `openedx.yaml` file, as described by OEP-2:
http://open-edx-proposals.readthedocs.io/en/latest/oeps/oep-0002.html

The data in this file was transformed from the contents of
edx/repo-tools-data:repos.yaml

### DIFF
--- a/openedx.yaml
+++ b/openedx.yaml
@@ -1,0 +1,8 @@
+# This file describes this Open edX repo, as described in OEP-2:
+# http://open-edx-proposals.readthedocs.io/en/latest/oeps/oep-0002.html#specification
+
+nick: prog
+oeps: {}
+openedx-release: {ref: master}
+owner: edx/ecommerce
+track-pulls: true


cpennington:
@edx/ecommerce: Marking this repo as one of yours.

openedx.yaml
```diff
@@ -0,0 +1,8 @@
+# This file describes this Open edX repo, as described in OEP-2:
+# http://open-edx-proposals.readthedocs.io/en/latest/oeps/oep-0002.html#specification
+
+nick: prog
```
rlucioni: programs

rlucioni:
@cpennington is there a preferred way to use this file to mark repos as having been deprecated? Programs is gradually being replaced by the catalog service.

openedx.yaml
cpennington: See the description of the field here: https://github.com/edx/open-edx-proposals/pull/18/files#diff-13c9b7b8133d4bcf6ab907c7f0065187R53

cpennington:
Yes, there is an `obsolete` tag (see the definition here: https://github.com/edx/open-edx-proposals/pull/18/files#diff-13c9b7b8133d4bcf6ab907c7f0065187R101)

rlucioni:
@cpennington good to know, thank you.

openedx.yaml
rlucioni: I see it, thanks. Why so short? Nothing else in our system refers to this service as "prog."

cpennington: Because there is particular Open edX tooling that builds reports that don't have much space for labels, I believe. @nedbat could probably give you more details.

rlucioni: I see, so these are used as labels on the axis of a graph or something? That might be worth adding to the description at https://github.com/edx/open-edx-proposals/pull/18/files#diff-13c9b7b8133d4bcf6ab907c7f0065187R53 for people who have the same question in the future.

rlucioni:
👍 once commits are squashed.

PR #166, opened 2016-09-29
Upgrade Django to 1.8.15

rlucioni:
@nedbat


rlucioni:
jshint failures are a surprise to me, I don't want to make changes apart from the Django upgrade here.

nedbat:
Any clue why jshint is failing? Is it something master has addressed?

rlucioni:
@nedbat that must be it, I merged a PR yesterday that passed all checks. Thoughts on what to do here?

nedbat:
Not sure why it used to pass, that line is clearly too long, but it's gone from master since 111be60c6c5b9ea68b53a3eb9697350dd8664eac.  This is good to merge. :+1:

PR #171, opened 2016-11-04
Use a 16.04 docker image to test

jibsheet:
This is not 100% yet - it doesn't pass down the token for codecov, I need to find where we set it and pass it into the docker container.

There are about 7 ways to skin the container build/run cat, but I think this is simple if a bit inelegant.
I'll rename the shell script, and Ed and I discussed possibly generating it from the travis.yml runtime instead.  

Just wanted to get any initial thoughts/concerns from @e0d @clintonb @rlucioni

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,18 +1,20 @@
 language: python
 python:
   - "2.7"
-sudo: false
+sudo: required
+services:
+  - docker
 
 cache:
   - pip
+
 before_install:
-  - "export DISPLAY=:99.0"
-  - "sh -e /etc/init.d/xvfb start"
-install:
-  - pip install -U pip wheel codecov
-  - make requirements
+  - docker-compose -f docker-compose-travis.yml up -d
+
 script:
-  - make validate_translations
-  - make validate
+  - docker exec programs_testing /edx/app/programs/programs/.travis/run_tests.sh
+
 after_success:
+  - pip install -U codecov
+  - docker exec programs_testing /edx/app/programs/programs/.travis/run_coverage.sh
   - codecov
--- a/.travis/run_coverage.sh
+++ b/.travis/run_coverage.sh
@@ -1,0 +1,4 @@
+#!/bin/bash -xe
+. /edx/app/programs/venvs/programs/bin/activate
+cd /edx/app/programs/programs
+coverage xml
--- a/.travis/run_tests.sh
+++ b/.travis/run_tests.sh
@@ -1,0 +1,7 @@
+#!/bin/bash -xe
+. /edx/app/programs/venvs/programs/bin/activate
+cd /edx/app/programs/programs
+pip install -U pip wheel
+make requirements
+make validate_translations
+make validate
--- a/docker-compose-travis.yml
+++ b/docker-compose-travis.yml
@@ -1,0 +1,16 @@
+# This is only currently used by Travis for testing
+version: "2"
+
+services:
+  programs:
+    image: edxops/programs:latest
+
+    container_name: programs_testing
+    volumes:
+      - .:/edx/app/programs/programs
+      - $HOME/.cache/pip:/edx/app/programs/.cache/pip
+    # The docker container produced from configuration.git does not currently
+    # run and expose a port for programs.  This means that we need to run some
+    # command that keeps the programs container alive while we run tests on it.
+    # We have not yet standardized on an init replacement which could be used instead.
+    command: tail -f /dev/null


run_tests.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash -xe
+apt-get update && apt-get install -y gettext
```
clintonb: Nit: do this as part of building the container.

run_tests.sh
```diff
@@ -0,0 +1,10 @@
...
+. /edx/app/programs/venvs/programs/bin/activate
+cd /edx/app/programs/programs
+pip install -U pip wheel codecov
```
clintonb: Nit: do this as part of building the container.

run_tests.sh
```diff
@@ -0,0 +1,10 @@
...
+make requirements
+make validate_translations
+make validate
+
+#codecov
```
clintonb: Have you seen https://github.com/codecov/support/wiki/Testing-with-Docker?

.travis.yml
```diff
@@ -1,18 +1,19 @@
...
+
 before_install:
-  - "export DISPLAY=:99.0"
-  - "sh -e /etc/init.d/xvfb start"
-install:
-  - pip install -U pip wheel codecov
-  - make requirements
+  - docker pull edxops/programs:latest
+  - docker run --name programs_testing -t -d -v $HOME/.cache/pip:/edx/app/programs/.cache/pip -v $PWD:/edx/app/programs/programs edxops/programs /bin/bash
+  - docker cp run_tests.sh programs_testing:/edx/app/programs/
```
clintonb: Shouldn't this be covered by the shared volume: `$PWD:/edx/app/programs/programs`?

.travis.yml
```diff
@@ -1,18 +1,19 @@
...
   - pip
+
 before_install:
-  - "export DISPLAY=:99.0"
-  - "sh -e /etc/init.d/xvfb start"
-install:
-  - pip install -U pip wheel codecov
-  - make requirements
+  - docker pull edxops/programs:latest
+  - docker run --name programs_testing -t -d -v $HOME/.cache/pip:/edx/app/programs/.cache/pip -v $PWD:/edx/app/programs/programs edxops/programs /bin/bash
```
clintonb: Can we use docker-compose to organize this? That would make it simpler if we decide to bring in more containers (e.g. MySQL, Elasticsearch) for more robust testing.

run_tests.sh
jibsheet: I thought about installing gettext in the docker container, but we build that with the ansible role.  I'm not 100% sure we want to ship gettext to production, although there may be little real harm?

Similar thought for codecov

run_tests.sh
jibsheet: I had not, I knew I needed to resolve it and wanted thoughts on the other parts.  Thanks for the pointers, it gave me enough info to figure it out in a secondary testing branch that I'll merge back here.

.travis.yml
jibsheet: Yeah, thought about that, was wondering if people would try to use programs in docker if I provided compose.  I suppose I can just document it's not for local dev.

This pattern is going to end up in other repos, so I was trying to contain my changes to .travis.yml when possible in order to not add extra weirdness for folks (or provide things someone might *think* should work for development but don't).

.travis.yml
jibsheet: You're correct, originally I was doing this differently, it's been removed.

jibsheet:
@rlucioni To your point, seems to add about 30 seconds, which I think is ok for getting actual 16.04 testing.  When they ship 16.04 images, or more native docker support, we can speed back up I hope.

To Both - thanks for looking, I'll be updating this with some extra work this afternoon.  I'm on call, so will likely have to let someone else from the team do the shipping, but I don't expect major changes to this branch.  I want to test that failures are detected properly, etc etc.

jibsheet:
@clintonb @rlucioni FYI - I've pushed what I think is the final version of this and will be merging it after tests pass unless you spot something you hate.

docker-compose.yml
```diff
@@ -0,0 +1,12 @@
...
+
+services:
+  programs:
+    image: edxops/programs:latest
+
+    container_name: programs_testing
+    volumes:
+      - .:/edx/app/programs/programs
+      - $HOME/.cache/pip:/edx/app/programs/.cache/pip
+    command: tail -f /dev/null # We're just keeping this alive as a place to run tests
```
rlucioni: @jibsheet can you elaborate on what's being done here?

rlucioni: Got it, thanks.

docker-compose.yml
```diff
@@ -0,0 +1,12 @@
+# This is only currently used by Travis for testing
```
clintonb: Can we rename this file to `docker-compose-travis.yml` to better reflect its usage, and not conflict (at least for other repositories) with the local development `docker-compose.yml`?

.travis.yml
```diff
@@ -1,18 +1,20 @@
...
 cache:
   - pip
+
 before_install:
-  - "export DISPLAY=:99.0"
-  - "sh -e /etc/init.d/xvfb start"
-install:
-  - pip install -U pip wheel codecov
-  - make requirements
+  - docker-compose up -d
```
clintonb: `docker-compose up -d -f docker-compose-travis.yml`

.travis.yml
```diff
@@ -1,18 +1,21 @@
...
   - pip
+
 before_install:
-  - "export DISPLAY=:99.0"
-  - "sh -e /etc/init.d/xvfb start"
-install:
-  - pip install -U pip wheel codecov
-  - make requirements
+  - docker-compose --version
+  - docker-compose up -d -f docker-compose-travis.yml
```
clintonb: Ordering may matter: `docker-compose -f docker-compose-travis.yml up -d`

https://docs.docker.com/compose/reference/

jibsheet: Yeah, took me a minute to parse their error.  Fixed now.

PR #187, opened 2016-12-09
Updated configuration for Docker-based devstack

clintonb:
ECOM-6570


clintonb:
@rlucioni @davec-edx @jmbowman please review
FYI @edx/devops

Makefile
```diff
@@ -14,6 +15,7 @@ help:
 	@echo "  help                       display this help message"
 	@echo "  html_coverage              generate and view HTML coverage report"
 	@echo "  migrate                    apply database migrations"
+	@echo "  production-requirements    install requirements for production"
```
davec-edx: use of - in target name...other targets use _

Makefile
```diff
@@ -32,7 +34,10 @@ clean:
 	coverage erase
 
 requirements:
-	pip install -r requirements/local.txt
+	pip install -r requirements/local.txt --exists-action w
+
+production-requirements:
```
davec-edx: - v _

programs/settings/base.py
```diff
@@ -1,4 +1,6 @@
 import os
+import platform
+from logging.handlers import SysLogHandler
 from os.path import join, abspath, dirname
```
davec-edx: I'm thinking we should have a common settings for IDAs or configuration service.  

Do we have a way of dynamically changing things, such as the logging level?

programs/settings/base.py
```diff
@@ -251,3 +253,76 @@
...
+            'propagate': True,
+            'level': 'WARNING'
+        },
+        '': {
+            'handlers': ['console', 'local'],
+            'level': 'DEBUG',
+            'propagate': False
+        },
+    }
+}
```
davec-edx: Consider tabular layout.

programs/settings/base.py
rlucioni: @davec-edx @jmbowman's work may help address this. We don't have a good answer right now.

requirements/production.txt
```diff
@@ -4,5 +4,6 @@
 gevent==1.0.2
 gunicorn==19.3.0
 MySQL-python==1.2.5
+newrelic==2.72.1.53
 PyYAML==3.11
 nodeenv==0.13.6
```
rlucioni: Nodeenv is still here. Shouldn't we be consistent about how we choose to install it?

programs/settings/base.py
rlucioni: @davec-edx let's address logging-related changes separately. As it stands, @clintonb's just moving what was already here.

programs/settings/base.py
clintonb: I agree that we should have common settings. At some point, I will repurpose https://github.com/edx/edx-django-extensions for this need.

requirements/production.txt
clintonb: Yes, it should go away here.

PR #193, opened 2017-01-22
Updated login/logout path

clintonb:
The paths are no longer nested under accounts. This brings the service in line with every other service.

ECOM-6570


clintonb:
FYI @edx/ecommerce

rlucioni:
Probably not worth making the change to use `from auth_backends.urls import auth_urlpatterns` at this point. 👍

PR #194, opened 2017-01-22
Pulling authentication URLs from auth_backends

clintonb:
- Removed duplication by replacing AUTH_URLS with values from auth_backends package
- Removed broken program path

ECOM-6570


clintonb:
FYI @rlucioni

rlucioni:
👍

PR #195, opened 2017-01-22
Removed unused context processor and setting

clintonb:
ECOM-6570


clintonb:
@edx/ecommerce please review

PR #203, opened 2017-03-06
Archive documentation

mikedikan:
ECOM-7205

Archive documentation for programs service.  Approach suggested by Ned Batchelder and is part of OEP-14.

@edx/ecommerce for review


docs/features.rst
```diff
@@ -1,3 +1,6 @@
+.. note:: This service is `being archived <https://groups.google.com/forum/#!topic/edx-code/Gn-qso9ZQNI>`_ and will no
```
clintonb: Usine [`include`](http://docutils.sourceforge.net/docs/ref/rst/directives.html#include) to avoid duplication.

The service *has been* archived and *is* no longer supported.

docs/archived.rst
```diff
@@ -0,0 +1,2 @@
+.. note:: This service has `been archived <https://groups.google.com/forum/#!topic/edx-code/Gn-qso9ZQNI>`_ and is no
+   longer be supported.
```
nedbat: "is no longer be supported" is no longer be English :)

rlucioni: s/longer be/longer

mikedikan: longer 🐝

mikedikan:
@clintonb Yup - I was just responding to the feedback given at retro to NOT squash commits until just before merge back to master so comments are preserved inline for commits.

PR #204, opened 2017-03-09
Archive the programs repository

mikedikan:
ECOM-7202

@edx/ecommerce for review and FYI


openedx.yaml
```diff
@@ -6,3 +6,4 @@ oeps: {}
 openedx-release: {ref: master}
 owner: edx/ecommerce
```
rlucioni: Could you remove this `owner` key? According to OEP-14, "it is not necessary for the openedx.yaml file to define an owner when archived is set to True."

mikedikan:
@rlucioni & @efagin - I read that as not necessary but also not discouraged - I wonder if we should use stronger language in the OEP

PR #207, opened 2017-12-21
LEARNER-3461 updated edx-i18n-tools version

Ayub-Khan:
Updated edx-i18n-tools version to latest.


ahsan-ul-haq:
There is no need to make any changes here as this IDA is [ARCHIVED]

